PR #1, opened 2017-02-20
Trying Gson for simpler implementation.

davidcarboni:
I've switched the CollectionInstrument entity to use Gson instead of Jackson, which makes the implementation less verbose. 

I thought I'd PR this to check we're happy with it.


davidcarboni:
Checked with Mark Price and agreed with Charles.

PR #2, opened 2017-02-22
Jdbc

davidcarboni:
I've been able to get Spring Boot to work with the Postgres Jsonb data type. If we're happy with this implementation for now, let's get this merged into master.

Thanks
David


martinlt:
I wasn't in the design session that you held and haven't seen the output, so with that said, I don't understand why we have decided to tightly coupled this code to Postgres and use the Jsonb datatype. What is the value of this? - it seems to have been more work than using two tables and JPA, means we lose the benefit of abstraction that JPA provides and if we switch to another SQL implementaton will mean more work (i.e. ONS may well shift from Postgres to SQL Server). 

Are there specific reasons for using this proprietary datatype?

PR #3, opened 2017-04-25
Us001 020 refactor ras-collection-instrument to align with ras-secure-message

duncannixon:


ras-collection-instrument/settings.py
```diff
@@ -0,0 +1,39 @@
...
+
+DEBUG = False
+TESTING = False
+CSRF_ENABLED = True
+SECRET_KEY = 'this-really-needs-to-be-changed'
+dbname = "ras_collection_instrument"
+SQLALCHEMY_DATABASE_URI = os.environ.get('SQLALCHEMY_DATABASE_URI', "postgresql://" + dbname + ":password@localhost:5431/postgres")
+
+
+#OAuthConfig
```
duncannixon: This comment was taken from settings.py in ras-secure-message, moved across into ras-ci to keep it consistent. My brief was to make the ras-ci service and ras-party handle their config in the same fashion as ras-secure-message.

I'm not precious about this but I would have thought that having a comment in there to describe the purpose of the file was good practice.

Perhaps it would be better to use the hash character for multiline comments as the codacy tool doesn't seem to like these multiline comments?

i.e. it doesn't seem to understand the last line of the comment:

All configuration may be overridden by setting the appropriate environment variable name. '''

LJBabbage: The original didn't have everything commented out as this one does. I assumed it was therefore redundant and should be removed

PR #8, opened 2017-05-05
Spike for Test Suite Structure

LJBabbage:
Not for merging, just for reveiw


nherriot:
We need at least 1 sunny day test case for each endpoint before we can merge this.
I have updated a README.md file and also added BDD lib in the requirements_test.tex file in my commit. This is so we can have one place to explain how all tests are done.

PR #11, opened 2017-05-25
Release/0.2 - merge of the demo repository

oddjobz:
This is an initial merge to get the new auto-generated code back into the main repo. The tests still need some refactoring, but currently the build should be passing with 100% coverage.


.gitignore
```diff
@@ -1,5 +1,68 @@
-.oracle_jre_usage/
-.idea/
-*.pyc
-ras-collection-instrument/uploads
+# Byte-compiled / optimized / DLL files
```
oddjobz: Ok, not sure I understand where this is coming from. The .gitignore is the standard template generated by GitHub for python applications, I've made three changes which are to add .idea, .build and .build test (??)

.travis.yml
```diff
@@ -1,16 +1,13 @@
...
-- CODECOV_TOKEN="d58091c4-cca2-4533-b419-bac51ed1511f"
-
 python:
-  - "2.7"
-
+  - "3.5"
 install:
-  - "pip install -r requirements.txt"
-
+  - pip install -r requirements_for_test.txt
```
oddjobz: Yup, removing the _for_test version. This is a hangover from the auto-generator, Jon has optimised to incorporate the requirements.txt as a component of the test- version, I didn't spot the name change when I refreshed.

config.ini
```diff
@@ -0,0 +1,25 @@
+[DEFAULT]
+db_driver = postgresql
+db_user = postgres
+db_pass = postgres
+db_port = 5432
+db_name = gareth
```
oddjobz: I would agree, but this is an auto-generated file, so I'll add this as a note on the codegen repo .. will change the default db name to "postgres" unless anyone has another preference.

requirements.txt
```diff
@@ -1,26 +1,32 @@
-alembic==0.9.1
 appdirs==1.4.3
```
oddjobz: Ditto, will add to comments on codegen.

requirements_for_test.txt
```diff
@@ -0,0 +1,35 @@
+appdirs==1.4.3
```
oddjobz: File removed.

scripts/test.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+if ! [ -a .build-test ] ; then
+	echo "Creating Virtual Environment"
+	virtualenv .build-test -p python3
+fi
+source .build-test/bin/activate
+pip3 -q install -r requirements_for_test.txt
+pytest --cov=swagger_server/controllers_local --cov-report term-missing
```
oddjobz: Mmm, they don't do the same thing. run_unit_test.sh is there for anyone working with 'workon' who wants html versions of reports, test.sh is there for virtenv users who just want cli information. Yes you could put the code in the same file and parameterise, but it's not something that would be done in codegen as these are auto-generated.

scripts/upload.py
```diff
@@ -0,0 +1,36 @@
+#!/usr/bin/env python3
```
oddjobz: Yes it is, indeed it's critical for testing.

swagger_server/controllers/ciupload_controller.py
```diff
@@ -0,0 +1,95 @@
+import connexion
```
oddjobz: Correct, all the routines currently in this auto-generated file have been overridden by live code.

swagger_server/controllers/ciupload_controller.py
```diff
@@ -0,0 +1,95 @@
...
+def activate_id_put(id):
+    """
+    Activate batch
+    Mark all items in the batch as available
+    :param id: Collection exercise identifier
+    :type id: str
+
+    :rtype: None
+    """
+    return 'do some magic!'
```
oddjobz: Not without recompiling the code generator and changing the templates we're using. We probably will get to doing this, but not at this stage.

swagger_server/configuration.py
```diff
@@ -0,0 +1,141 @@
...
+    def _activate_database(self):
+        """
+        Connect to the database (create it if it's missing) and set up tables as per our models.
+        If we're in a 'test' environment, drop all the tables first ...
+        """
+        self._session.remove()
+        self._session.configure(bind=self._engine, autoflush=False, autocommit=False, expire_on_commit=False)
+        if not database_exists(self._engine.url):
+            create_database(self._engine.url)
+        if self.if_drop_database:
```
oddjobz: More auto-generated code, this has since been refactored by Jon, I will run swagger-referesh presently and re-push.

swagger_server/configuration.py
```diff
@@ -0,0 +1,141 @@
...
+            self._base.metadata.drop_all(self._engine)
+        from .models_local import _models
```
oddjobz: If we set "drop_database" for a given profile, it drops all tables in the database as part of the startup procedure. This is used by the "test" profile to ensure testing starts with a clean database. The import references the _models.py file in /models_local which loads all Alchemy models in on the defined declarative_base(), such that the next call to create_all recreates any missing tables.

swagger_server/configuration.py
```diff
@@ -0,0 +1,141 @@
...
+        self._base.metadata.create_all(self._engine)
+
+    def _activate_cf(self):
+        """
+        If we're running on Cloud Foundry we need to set the host in the swagger YAML file to be
+        the hostname we recover from VCAP application. We also need to set the configuration port for
+        use by the listener if we're running locally.
+        """
+        self._crypto_key = getenv('ONS_CRYPTOKEY', self.get('crypto_key'))
+        config = './swagger_server/swagger/swagger.yaml'
```
oddjobz: Yes it could go in config.ini, but the location is set by the code generator, i.e. it's a constant and config.ini is designed to track things that could change so they can be altered without altering the code. If we do this there are a few other instances that should also be refactored, but again this is all codegen stuff and not project specific. Would be good to get issues logged on the Codegen repo ..

setup.py
```diff
@@ -0,0 +1,33 @@
+# coding: utf-8
+
+import sys
```
oddjobz: Auto-generated, will log an issue on the codegen repo.

swagger_server/controllers/respondent_controller.py
```diff
@@ -0,0 +1,34 @@
+import connexion
```
oddjobz: controllers is auto-generated and overridden. Without recompiling the java app and changing the templates we can't do much about this.

swagger_server/controllers_local/respondent_controller.py
```diff
@@ -0,0 +1,50 @@
...
+#   Collection Instruments Upload                                            #
+#   Date:    11 May 2017                                                     #
+#   Author:  Gareth Bult                                                     #
+#   License: MIT                                                             #
+#   Copyright (c) 2017 Crown Copyright (Office for National Statistics)      #
+#                                                                            #
+##############################################################################
+
+import os
+from flask import request, jsonify, make_response
```
oddjobz: removed

swagger_server/models/__init__.py
```diff
@@ -0,0 +1,8 @@
+# coding: utf-8
```
oddjobz: Autogenerated.

swagger_server/models/collectioninstrument.py
```diff
@@ -0,0 +1,153 @@
...
+from typing import List, Dict
+from ..util import deserialize_model
+
+
+class Collectioninstrument(Model):
+    """
+    NOTE: This class is auto generated by the swagger code generator program.
+    Do not edit the class manually.
+    """
+    def __init__(self, classifiers: Classifiers=None, survey_id: str='urn:ons.gov.uk:id:survey:001.234.56789', ci_type: str='OFFLINE', id: str='urn:ons.gov.uk:id:ci:001.234.56789'):
```
oddjobz: This comes from the original API specification, this does need a tidy up, but it's not in-scope for the moment. (it's auto-generated and overridden)

swagger_server/models_local/_models.py
```diff
@@ -0,0 +1,6 @@
+from .instrument import InstrumentModel
```
oddjobz: This needs to be there as above, used by configuration.py.

swagger_server/models_local/business.py
```diff
@@ -0,0 +1,24 @@
+from sqlalchemy import Column
+from sqlalchemy.orm import relationship
+from sqlalchemy.types import Integer, Enum, String
```
oddjobz: Removed

swagger_server/models_local/business.py
```diff
@@ -0,0 +1,24 @@
...
+from ..configuration import ons_env
+from .associations import instrument_business_table
+from .guid import GUID
```
oddjobz: Removed

swagger_server/models_local/instrument.py
```diff
@@ -0,0 +1,62 @@
+from sqlalchemy import Column, ForeignKey
+from sqlalchemy.orm import relationship
+from sqlalchemy.types import TIMESTAMP, String, Integer, LargeBinary
```
oddjobz: Removed

swagger_server/models_local/survey.py
```diff
@@ -0,0 +1,22 @@
+from sqlalchemy import Column
```
oddjobz: Removed

swagger_server/test/test_ciupload_controller.py
```diff
@@ -0,0 +1,90 @@
+# coding: utf-8
```
oddjobz: Autogenerated.

swagger_server/test/test_respondent_controller.py
```diff
@@ -0,0 +1,41 @@
+# coding: utf-8
+
+from __future__ import absolute_import
```
oddjobz: Autogenerated.

swagger_server/test/test_static_controller.py
```diff
@@ -0,0 +1,26 @@
+# coding: utf-8
+
+from __future__ import absolute_import
+
+from . import BaseTestCase
+from six import BytesIO
+from flask import json
```
oddjobz: Autogenerated

swagger_server/controllers_local/collectioninstrument.py
```diff
@@ -0,0 +1,300 @@
+##############################################################################
```
oddjobz: An example would be useful (?)

swagger_server/controllers_local/collectioninstrument.py
```diff
@@ -0,0 +1,300 @@
...
+from ..models_local.exercise import ExerciseModel
+from ..models_local.business import BusinessModel
+from ..models_local.survey import SurveyModel
+from ..models_local.classification import ClassificationModel, classifications
+from traceback import print_exc
+from sys import stdout
+from json import loads
+from uuid import UUID
+
+DEFAULT_SURVEY = "3decb89c-c5f5-41b8-9e74-5033395d247e"
```
oddjobz: Yes, this is the UUID of "BRES". This is needed for current uploads that do not provide a uid for the survey in use, yet we need to associate uploads with a given survey (uid). This will change once we know more about how we will interact with other uploads. It might be good to have an application specific ini file to put such constants in, but that infrastructure doesn't exist yet.

swagger_server/controllers_local/collectioninstrument.py
```diff
@@ -0,0 +1,300 @@
...
+    through by mistake. It then converts the first parameter to a UUID type if the 'uuid' flag is set.
+    
+    :param uuid: Whether to convert first parameter to UUID() 
+    :return: status code, return message
+    """
+    def protect_wrapped(func):
+        def func_wrapped(*args, **kwargs):
+            try:
+                my_id = args[1]
+                if my_id == 'id_example':
```
oddjobz: Neither am I. I'm open to suggestions, but this is to interact with the auto-generated tests - if you have a better mechanism I'm all ears .. for now, this allows us to use the auto-generated integration tests.

swagger_server/controllers_local/collectioninstrument.py
```diff
@@ -0,0 +1,300 @@
...
+
+    """
+    Database shortcuts, the SQLAlchemy syntax isn't always immediately obvious, so here we're just using
+    some 'more' obvious shortcuts to functions, purely from a 'readability' perspective.
+    """
+    def _get_exercise(self, exercise_id):
+        return ons_env.session.query(ExerciseModel).filter(ExerciseModel.exercise_id == exercise_id).first()
+
+    def _get_business(self, ru_ref):
+        return ons_env.session.query(BusinessModel).filter(BusinessModel.ru_ref == ru_ref).first()
```
oddjobz: Absolutely. Errors are picked up by "protect". Database specific exceptions can be added here once we know sorts of errors are likely where a specific response or action is appropriate.

swagger_server/test_local/test_ciupload_controller.py
```diff
@@ -0,0 +1,313 @@
...
+        """
+        code, msg = self.collection_instrument.define_batch(str(uuid4()), 100)
+        self.assertTrue(code == 200, msg)
+        try:
+            code, msg = self.collection_instrument.define_batch(0, 100)
+            self.assertTrue(code == 200, msg)
+        except:
+            pass
+
+    def test_03_upload(self):
```
oddjobz: ??

swagger_server/test_local/test_ciupload_controller.py
```diff
@@ -0,0 +1,313 @@
...
+            fileobject = FileStorage(stream=io, filename='dummy.txt', name='myname')
+            code, msg = self.collection_instrument.upload(str(uuid4()), fileobject, 'dummy.txt')
+            self.assertTrue(code == 200, msg)
+
+        with open('scripts/upload.txt', 'rb') as io:
+            fileobject = FileStorage(stream=io, filename='dummy.txt', name='myname')
+            code, msg = self.collection_instrument.upload(0, fileobject, 'dummy.txt')
+            self.assertTrue(code == 400, msg)
+
+    def test_04_csv(self):
```
oddjobz: Numbering ensures the tests are run in order rather than randomly .. when developing tests and watching output it is **very** helpful, at the very lest to me. Test functions are all commented, why do we need more descriptive function names?

swagger_server/test_local/test_ciupload_controller.py
```diff
@@ -0,0 +1,313 @@
...
+        """
+        Upload a file, then make sure we are able to download the result in CSV format
+        """
+        batch = str(uuid4())
+        with open('scripts/upload.txt', 'rb') as io:
+            fileobject = FileStorage(stream=io, filename='dummy.txt', name='myname')
+            #self.assertTrue(self.collection_instrument.upload(batch, fileobject, 'dummy.txt'))
```
oddjobz: Hangover from last refactoring pass - removed.

swagger_server/test_local/test_ciupload_controller.py
```diff
@@ -0,0 +1,313 @@
...
+            code, msg = self.collection_instrument.upload(batch, fileobject, 'dummy.txt')
+            self.assertTrue(code == 200, msg)
+
+        code, msgs = self.collection_instrument.csv(batch)
+        msg = msgs.split('\n')[1][:21]
+        print(">>", msg)
```
oddjobz: Mmm, again a hangover from debugging I was doing, however worth noting that if a test fails you do a see these prints (!) ... have now removed all debugging print statements.

swagger_server/test_local/test_ciupload_controller.py
```diff
@@ -0,0 +1,313 @@
...
+        wanted = '"1","dummy.txt","24",'
+        print(">>", wanted)
+        self.assertTrue(msg == wanted, "CSV download")
+        code, msg = self.collection_instrument.csv(str(uuid4()))
+        self.assertTrue(code == 204, "CSV download")
+        try:
+            code, msg = self.collection_instrument.csv(0)
+        except Exception:
+            pass
```
oddjobz: Yup, status check missing, have added. It will always fail, it's only really there for coverage.

swagger_server/test_local/test_ciupload_controller.py
```diff
@@ -0,0 +1,313 @@
...
+        print("Code=", code)
+        self.assertTrue(code == 400, 'Failed to lookup instrument with integer id')
+
+        try:
+            code, msg = self.collection_instrument.instrument()
+        except Exception:
+            pass
+        print("Code=", code)
```
oddjobz: It will *always* return a code, the next line down should check / catch the correct status

swagger_server/test_local/test_ciupload_controller.py
```diff
@@ -0,0 +1,313 @@
...
+        self.assertTrue(code == 400, msg)
+
+    def test_11_instruments(self):
+        """
+        Try out some of the query by classifier options
+        """
+        batch = str(uuid4())
+        with open('scripts/upload.txt', 'rb') as io:
+            fileobject = FileStorage(stream=io, filename='dummy.txt', name='myname')
+            #self.assertTrue(self.collection_instrument.upload(batch, fileobject, 'file1'))
```
oddjobz: Commented code all removed.

swagger_server/test_local/test_static_controller.py
```diff
@@ -0,0 +1,25 @@
+# coding: utf-8
```
oddjobz: Gone ...

swagger_server/controllers_local/collectioninstrument.py
LJBabbage: I see the same
```
exercise = self._get_exercise(exercise_id)
        if not exercise:
            return 204, {'text': 'No such batch'}
```
At worst the json result should be extracted for maintainability. This would also apply for {'text': 'OK'}

swagger_server/controllers_local/collectioninstrument.py
LJBabbage: If someone passed 'id_example' as an arg outside of a unit test what happens? Happy for the moment whilst we are working with the auto generated tests, but we should be testing properly, with proper ids

swagger_server/controllers_local/collectioninstrument.py
LJBabbage: So the whole of the code in collectioninstrument.py is effectively in a try/except? And 'def instruments'  is a nested try/except?

swagger_server/controllers_local/collectioninstrument.py
```diff
@@ -0,0 +1,300 @@
...
+        exercise = self._get_exercise(exercise_id)
+        if not exercise:
+            return 204, {'text': 'No such batch'}
+        if exercise.status != 'pending':
+            return 500, {'text': 'Batch in wrong state'}
+        exercise.status = 'active'
+        ons_env.session.commit()
+        return 200, exercise.json
+
+    @protect(uuid=True)
```
LJBabbage: You don't necessarily need to set these all to true as the default will handle this

swagger_server/test_local/test_ciupload_controller.py
LJBabbage: Test_03 is testing 2 things, a valid upload and an upload to an undefined batch, I would expect 2 separate tests

swagger_server/test_local/test_ciupload_controller.py
LJBabbage: I have always be taught that tests should be able to run independently. I think the randomness of tests is a good thing and avoids dependencies hidden by a potential order. However I am happy to defer to a team decision on this one. 

Having good naming policies on variables and functions regardless of situation is my preferred option, but if the team and lead feel it is unnecessary I will defer

swagger_server/configuration.py
jonmorphic: Agree GitHub issues are a good way to track TODO items for the code generator.

swagger_server/controllers_local/respondent_controller.py
```diff
@@ -0,0 +1,50 @@
...
+#   License: MIT                                                             #
+#   Copyright (c) 2017 Crown Copyright (Office for National Statistics)      #
+#                                                                            #
+##############################################################################
+
+import os
+from flask import jsonify, make_response
+from .collectioninstrument import CollectionInstrument
+
+root_folder = os.getcwd()
```
jonmorphic: Don't think root_folder is used?

swagger_server/test_local/test_ciupload_controller.py
```diff
@@ -0,0 +1,294 @@
...
+    def test_01_define_batch(self):
+        """
+        Make sure we can define a batch
+        """
+        code, msg = self.collection_instrument.define_batch(str(uuid4()), 100)
+        self.assertTrue(code == 200, msg)
+        try:
+            code, msg = self.collection_instrument.define_batch(0, 100)
+            self.assertTrue(code == 200, msg)
+        except:
```
jonmorphic: Is the code-under-test expected to throw an exception here?

swagger_server/test_local/test_ciupload_controller.py
jonmorphic: I'd agree with this: Tests should ideally be order-independent and idempotent. Long descriptive test names helps make the tests self-documenting.

swagger_server/test_local/test_ciupload_controller.py
jonmorphic: Does it always fail by specification? In which case should the test explicitly check that an exception is thrown?

swagger_server/controllers_local/collectioninstrument.py
oddjobz: Sure, when we have some standard for this / when we know what the caller wants in terms of textual response, we can standardise. Not really worth fixing until then as it will change.

swagger_server/controllers_local/collectioninstrument.py
oddjobz: The code would (could) never do this as id's are validated, if someone was hacking an endpoint, it would give them a 200 instead of 204 .. I'm not too worried about a hacker being given the wrong status code .. indeed it might frustrate him for a while .. :)

swagger_server/controllers_local/collectioninstrument.py
oddjobz: Each endpoint is in a try / except ...

swagger_server/test_local/test_ciupload_controller.py
oddjobz: It's testing the upload routine in two different ways, why would you expect two tests?

swagger_server/test_local/test_ciupload_controller.py
oddjobz: Have removed all numbers and increased the length of test names. I'm not going to push any more to this branch as I'm working on another, so these fixes will be included in the next PR.

swagger_server/test_local/test_ciupload_controller.py
oddjobz: Have refactored this issue away - again, will be in next PR.

swagger_server/controllers_local/collectioninstrument.py
```diff
@@ -0,0 +1,300 @@
...
+    def instrument(self, instrument_id):
+        """
+        Return a specific collection instrument based on an id
+
+        :param instrument_id: The id of the instrument we want 
+        :return: formatted JSON version of the instrument
+        """
+        instrument = self._get_instrument(instrument_id)
+        if not instrument:
+            return 404, 'Instrument not found'
```
oddjobz: Good question - I don't believe we have an authoratative answer, the 204 was initially suggested by Duncan ..

swagger_server/controllers_local/collectioninstrument.py
LJBabbage: No problem

LJBabbage:
As agreed in this instance, fixes can be made in different PR

PR #12, opened 2017-05-31
TBL-041: Add JWT endpoint authorization

oddjobz:
Also incorporates tidyups from previous PR


requirements.txt
```diff
@@ -31,3 +31,4 @@ tox==2.7.0
 Twisted==16.6.0
 Werkzeug==0.12.1
 zope.interface==4.4.0
+python-jose==1.3.2
```
LJBabbage: Couldn't see this being used

swagger_server/configuration.py
```diff
@@ -180,6 +180,10 @@ def if_drop_database(self):
         return drop.lower() in ['yes', 'true']
 
     @property
+    def is_secure(self):
+        return self.get('authentication', 'true').lower() in ['yes', 'true']
```
LJBabbage: Would a default of false be better, just as it authentication?

PR #14, opened 2017-06-12
First pass at structured logging to JSON (eQ and SDX)

liamtoozer:
Added initial use of structlog (taken from eQ and SDX) to enable JSON log output.

### How to review 
Start the server (see README) and ensure a request to an end point shows output in JSON format. E.g.
`curl -X GET --header 'Accept: application/json' 'http://localhost:8080 /collection-instrument-api/1.0.2/collectioninstrument?searchString=%7B%22SIZE%22%3A%20%2225%22%7D'`

### DIFF
--- a/swagger_server/controllers_local/controller_helper.py
+++ b/swagger_server/controllers_local/controller_helper.py
@@ -1,0 +1,18 @@
+from uuid import uuid4
+from structlog import get_logger
+
+logger = get_logger()
+
+
+def ensure_log_on_error(code, msg):
+    if code != 200:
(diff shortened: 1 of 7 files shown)


swagger_server/controllers_local/controller_helper.py
```diff
@@ -0,0 +1,18 @@
+from uuid import uuid4
+from structlog import get_logger
+
+logger = get_logger()
+
+
+def ensure_log_on_error(code, msg):
```
LJBabbage: This needs to be expanded, any error being recorded as info will be lost if we ever needed to debug

PR #15, opened 2017-06-12
Collection-instruments: Respondent upload

LJBabbage:
To add 2 new end points from an update swagger definition (1.0.3), one to upload to a new model in the database and one to view the updated details. Validation is run against the file to check file type and file name length. 

### How to review 
Check the tests run successfully and hit the end points and make sure the files gets upload and that when you query the database it returns the expected record

N.B session.commit() and the corresponding flask error register has not been tested


oddjobz:
Ok, please don't merge this on as it's going to conflict with some other work. I'm in the process of merging these changes with the other changes and will submit a consolidated pull request ...

PR #16, opened 2017-06-15
Us034b/search interface

oddjobz:
Few stories here;

a. Refactor to use ons_ras_common rather than using in-line code (this is the vast majority of change)
b. Modify search routine for Titchfield (minor formatting changes)
c. Rework /download endpoint to work with instrumentId rather than ru_ref


config.ini
```diff
@@ -1,32 +1,108 @@
...
+;   5. int  - is the CF int environment
+;   6. prod - is the CF production environment
+;
 [test]
-db_connection = sqlite:///${db_test}
-db_drop = true
+drop_database = yes
+db_connection = sqlite:///${db_name}
 debug = true
 log_level = DEBUG
```
nherriot: add authentication = false

swagger_server/controllers_local/ciupload_controller.py
```diff
@@ -146,13 +146,13 @@ def download_id_get(id):
     """
     Download a file based on the id (RU_REF)
```
nherriot: comment to be updated

PR #17, opened 2017-06-16
Us032a/us034b/merge

oddjobz:
Ok, release 0.2 is a clone of master, so this is to merge Lewis's amended changes into 0.2. Once this is done, we can revert yesterday's PR against master.


oddjobz:
Scrapping

PR #27, opened 2017-07-17
Upload, check, encrpyt and send to rabbitmq

LJBabbage:
This is an early look at the rabbitmq process. At present we don't have the external queue or a key management process and the JWT comparisons has not be implemented. However everything else can be reviewed.

You will need to export the following environmental values and bind to a rabbit service through cf
API_GATEWAY_CASE_URL
API_GATEWAY_COLLECTION_EXERCISE_URL
PRIVATE_SIGNING_KEY
PRIVATE_SIGNING_KEY_PASSWORD
RABBITMQ_AMQP

### How to review 
Run the tests and application. A good way to use it is push it up to cf and then create a simple html to post to it like

```
<!DOCTYPE html>
<html lang="en">
<head>
    <meta charset="UTF-8">
    <title>Title</title>
</head>
<body>

<form action="http://collection-instrument-demo.apps.devtest.onsclofo.uk/collection-instrument-api/1.0.2/survey_responses/ab548d78-c2f1-400f-9899-79d944b87300" method="POST" enctype="multipart/form-data">
    <div>
        <label class="label" name="file">Select a file: </label>
        <input  name="file" type="file" />

    </div>

  <input type="submit" value="Submit">
</form>
</body>
</html>
```


jonmorphic:
Initial comment, I had some trouble running the unit tests, either via tox or test.sh. 

test.sh is fixed by deleting the .build-test directory, which seemed to be preventing correct installation of ons_ras_common.

tox is fixed by passing ONS_ENV=travis into the test runner (as with test.sh). This seems confusing, we're not running the tests on travis and it would be better to avoid having unit tests which are sensitive to the environment in that way.

Procfile
```diff
@@ -1 +1 @@
-web: python3 -m swagger_server
+web: ONS_ENV=development python3 -m swagger_server
```
jonmorphic: Is ONS_ENV=development correct for all "environments"?

manifest.yml
```diff
@@ -4,5 +4,6 @@ applications:
   memory: 512M
   host: collection-instrument-demo
```
jonmorphic: Is the hostname correct here?

manifest.yml
```diff
@@ -4,5 +4,6 @@ applications:
...
   services:
-    - ras-postgres
+    - rm-pg-db
```
jonmorphic: What's rm-pg-db?

requirements.txt
```diff
@@ -1,4 +1,4 @@
-git+https://github.com/ONSdigital/ras-common.git@release/0.3#ons_ras_common==0.1.90
+git+https://github.com/ONSdigital/ras-common.git@release/0.3#ons_ras_common==0.1.92"
```
jonmorphic: Typo here?

jonmorphic: Also, is release/0.3 the correct branch from which to consume ons_ras_common? I can see four branches of ras-common in flight at time of this comment. It may be acceptable as this PR is still in progress, but at point of merging it's problematic to consume a library off a branch.

swagger_server/controllers/cryptography/keys.py
```diff
@@ -0,0 +1,89 @@
+TEST_PRIVATE_SIGNING_KEY = """-----BEGIN ENCRYPTED PRIVATE KEY-----
```
jonmorphic: Are these test fixtures?

swagger_server/controllers/helper.py
```diff
@@ -1,8 +1,69 @@
...
+
+
+def is_valid_file_extension(file_name, extensions):
+    """
+    Check the file format is valid
+    :param file_name: The file name to be checked
+    :param extensions: The list of extensions that are valid
+    :return: boolean
+    """
     extensions_tuple = tuple([e for e in extensions.split(",")])
```
jonmorphic: Minor point: there's no need for the list comprehension here.

swagger_server/controllers/submitter/encrypter.py
```diff
@@ -0,0 +1,78 @@
...
+    TEST_SDX_PUBLIC_KEY
+KID = 'EDCSR'
+
+
+class Encrypter(JWEEncrypter):
+
+    # password and key variables are dynamically assigned
+    def __init__(self):
+
+        private_key = os.environ.get('PRIVATE_SIGNING_KEY', TEST_PRIVATE_SIGNING_KEY)
```
jonmorphic: This might be a good place for dependency injection, in order to make the class a bit simpler to test, and to allow the environmental dependencies to be made explicit somewhere visible.

swagger_server/controllers/submitter/encrypter.py
```diff
@@ -0,0 +1,78 @@
...
+        private_key_password = os.environ.get('PRIVATE_SIGNING_KEY_PASSWORD', TEST_PRIVATE_SIGNING_KEY_PASSWORD)
+        public_key = os.environ.get('SDX_PUBLIC_KEY', TEST_SDX_PUBLIC_KEY)
+
+        self._load_keys(private_key, private_key_password, public_key)
+
+        # first generate a random key
+        self.cek = os.urandom(32)  # 256 bit random CEK
```
jonmorphic: I'm guessing this is monkey-patched in the unit tests? Might be somewhere else to consider DI.

swagger_server/controllers/survey_response.py
```diff
@@ -25,58 +24,199 @@ class SurveyResponse(object):
...
+            # Create, encrypt and send message to rabbitmq
+            json_message = self._create_json_message_for_file(file, file_extension, period, ru, survey_id)
+            encrypted_message = self._encrypt_message(json_message)
+
+            rabbit_mq = RabbitMQSubmitter()
+
+            if rabbit_mq.send_message(encrypted_message, QUEUE_NAME, tx_id):
+                return 200, UPLOAD_SUCCESSFUL
+            else:
+                return 400, UPLOAD_UNSUCCESSFUL
```
jonmorphic: Should this be a 500?

swagger_server/controllers/submitter/rabbitmq_submitter.py
```diff
@@ -0,0 +1,57 @@
...
+            channel = self.connection.channel()
+            channel.queue_declare(queue=queue)
+            published = channel.basic_publish(exchange='',
+                                              routing_key=queue,
+                                              body=message,
+                                              properties=properties)
+            return published
+        except AMQPError:
+            ons_env.logger.error("Rabbitmq message not sent")
+            return False
```
jonmorphic: The recommended pattern in Python is for error paths to return via exception, rather than return a falsey value.

swagger_server/test/test_rabbitmq_submitter.py
```diff
@@ -0,0 +1,46 @@
...
+            rabbitmq_submitter = RabbitMQSubmitter()
+            rabbitmq_submitter._get_rabbitmq_url = self.mock_get_rabbitmq_url
+
+            # When a message is send
+            published = rabbitmq_submitter.send_message(message={}, queue='Seft.Responses', tx_id='test')
+
+            # Then it is published successfully
+            self.assertTrue(published)
+
+    def test_when_message_sent_then_published_fasle(self):
```
jonmorphic: good test  👍

manifest.yml
LJBabbage: The name of the service had changed on cf so I matched it

requirements.txt
LJBabbage: True, hopefully we can get back to just master soon

swagger_server/controllers/cryptography/keys.py
LJBabbage: Yep, I had to stick something in there whilst we have no plan on key management. They will be removed soon hopefully

swagger_server/controllers/helper.py
LJBabbage: Forgot about this one. The ons.env.get is returning an unusable string "xls,xlsx" which natural wouldn't work, so I put this in to get around it. I will have another look

manifest.yml
jonmorphic: If it's the name of the CF database service, then it should be unique to ras-collection-instrument, e.g. ras-ci-db

swagger_server/controllers/helper.py
jonmorphic: Just do `extensions.split(",")`, the comprehension isn't needed

swagger_server/controllers/submitter/encrypter.py
LJBabbage: I agree, but I will wait till I have full key management clarification before doing anything

LJBabbage:
Re - Testing. We definitely need to look at this as, I didn't even notice what it was doing. Its also ducking JWT authentication, which we need to stop

pricem14pc:
ALL GOOD

PR #29, opened 2017-08-09
Tblxxx/cfenv

oddjobz:
Fixes for /info and environment for prod and preprod.


pricem14pc:
Ok, I can't get /info working but we will need to take this into TEST/SIT/PREPROD as a defect.

PR #31, opened 2017-08-17
Return case id to seft consumer

tej99:


tej99:
[[SEFT] consumer receipting (5)](https://trello.com/c/3gFiWZ1v/250-seft-consumer-receipting-5)

PR #34, opened 2017-08-25
WIP: DON'T Merge File upload session management

warrenbailey:
Let the file upload process manage its own session as the data it puts in is massive and is filling the db cache

### DIFF
--- a/swagger_server/controllers/collectioninstrument.py
+++ b/swagger_server/controllers/collectioninstrument.py
@@ -313,7 +313,10 @@ class CollectionInstrument(object):
         logger.info('Uploading Ru-Ref: {}'.format(ru_ref))
         try:
             logger.info("Creating db models")
-            session = scoped_session(sessionmaker())
+            session_factory = sessionmaker(bind=ons_env.db.engine)
+            Session = scoped_session(session_factory)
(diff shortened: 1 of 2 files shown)


jonmorphic:
As far as I can tell, this change doesn't alter the session scope. My understanding of the likely problem is that the WSGI container accepts many concurrent incoming requests, thus opening multiple large SQLAlchemy sessions and blowing the CF app memory limit. If so, then we'd need to serialise the request handling, which is quite a significant change.

Have we been able to test this change somewhere?

warrenbailey:
Testing it now.

I think each Thread has its own Session is reusing that for every request and never closing it. The cache then builds up and eventually consumes all the memory.

In the transaction manager here, it's calling self._session which is a single instance of scoped session
https://github.com/ONSdigital/ras-common/blob/master/ons_ras_common/ons_database.py

jonmorphic:
Thanks for explaining. I’m seeing that the context manager always closes a session and the teardown handler removes it from the registry, however I’m not seeing where ras-common properly scopes the session so I think you could be right about that. I’d agree it’s worth testing to see the effect. We can discuss further on Tuesday. Thanks for helping out!

PR #38, opened 2017-09-01
Defect 26 unfiltered file upload

LJBabbage:
To make sure someone isn't trying to land exe files on ONS estate. The code makes use of the python-magic library

### How to review 
Run the test and then on cf try to upload an exe with an xls extension

### DIFF
--- a/swagger_server/controllers/survey_response.py
+++ b/swagger_server/controllers/survey_response.py
@@ -10,6 +10,7 @@ from swagger_server.controllers.helper import is_valid_file_extension, is_valid_
 from swagger_server.controllers.submitter.encrypter import Encrypter
 from swagger_server.controllers.submitter.rabbitmq_submitter import RabbitMQSubmitter
 from swagger_server.controllers.exceptions import UploadException
+from werkzeug.utils import secure_filename
 
 FILE_EXTENSION_ERROR = 'un-accepted file extension'
 FILE_NAME_LENGTH_ERROR = 'The name of the file is too long'
@@ -36,7 +37,7 @@ class SurveyResponse(object):
         ons_env.logger.info('Adding survey response')
 
         if case_id and file and hasattr(file, 'filename'):
-            file_name, file_extension = os.path.splitext(file.filename)
+            file_name, file_extension = os.path.splitext(secure_filename(file.filename))
             file_contents = file.read()
             is_valid_file, msg = self._is_valid_file(file_contents, file_name, file_extension)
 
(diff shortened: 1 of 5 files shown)


swagger_server/controllers/survey_response.py
```diff
@@ -35,7 +37,8 @@ def add_survey_response(self, case_id, file):
 
         if case_id and file and hasattr(file, 'filename'):
             file_name, file_extension = os.path.splitext(file.filename)
-            is_valid_file, msg = self._is_valid_file(file_name, file_extension)
+            file_contents = file.read()
```
jonmorphic: Is there are a risk here of reading large files into memory? I think to extract the magic number from the file, we may be able to read the first few bytes.

pricem14pc:
Why aren't we simply checking the mime type? Current RRM classic checks for application/vnd.openxmlformats-officedocument.spreadsheetml.sheet or application/vnd.ms-excel

How do we know that magic.from_buffer(file_contents) reliably resolves all xls / xlsx files (created with all client applications/platforms) to 'Composite Document File V2 Document'?

Does the message 'The file extension has been altered' refer to the respondent having  uploaded a different survey response file type to the collection instrument file type they downloaded? This may well be possible. They may have saved their survey response as an xls instead of xlsx. This would be a valid acceptable scenario. Or is this message misleading?

swagger_server/test/test_survey_response.py
```diff
@@ -144,6 +145,18 @@ def test_add_survey_response_long_file_name(self):
         self.assertEquals(status, 400)
         self.assertEquals(FILE_NAME_LENGTH_ERROR, msg)
 
+    def test_add_survey_response_file_extension_changed(self):
```
jonmorphic: 👍

swagger_server/controllers/survey_response.py
jonmorphic: We probably want to use secure_filename here also. See http://werkzeug.pocoo.org/docs/0.12/utils/

jonmorphic:
I think the mime type gets set on the client system, the exact algorithm varies, but generally based off file extension. Checking the magic number server side is probably safer.

LJBabbage:
The mime time can be easily manipulated, we need to check the file

swagger_server/controllers/survey_response.py
LJBabbage: Done

LJBabbage: We have to use the full file later

pricem14pc:
Approving but will be speaking to Dan O'Brien re: IA requirements.
Dependent on OS linux libmagic

PR #40, opened 2017-09-05
Isu 8 filename correction from a respondent upload

nherriot:
this fixes the file name format issue so that files are of the type:
60218763103X_201706_221_20170905113039.xls
there is also a new test case for it, some tidy up and another small code change fix where a variable was being set but never being used.


swagger_server/test/test_survey_response.py
```diff
@@ -49,6 +49,30 @@ def test_add_survey_response_success(self):
         self.assertEquals(status, 200)
         self.assertEquals(UPLOAD_SUCCESSFUL, msg)
 
+    def test_add_survey_response_success_check_file_name(self):
```
jonmorphic: 👍

PR #44, opened 2017-09-12
Isu 8 filename correction from a respondent upload

nherriot:
Fixes the file name being wrong.

### DIFF
--- a/swagger_server/controllers/survey_response.py
+++ b/swagger_server/controllers/survey_response.py
@@ -57,6 +57,7 @@ class SurveyResponse(object):
             if collection_exercise:
                 exercise_ref = collection_exercise.get('exerciseRef')
                 survey_id = collection_exercise.get('surveyId')
+                ons_env.logger.debug('generating file name for upload with a survey ID of: {}'.format(survey_id))
             else:
                 return self._invalid_upload()
 
@@ -219,7 +220,7 @@ class SurveyResponse(object):
 
         check_letter = 'X'
         time_date_stamp = time.strftime("%Y%m%d%H%M%S")
-        file_name = "{ru}{check_letter}_{exercise_ref}_{survey_id}_{time_date_stamp}.{file_format}".format(ru=ru,
+        file_name = "{ru}{check_letter}_{exercise_ref}_{survey_id}_{time_date_stamp}{file_format}".format(ru=ru,
                                                                                 check_letter=check_letter,
                                                                                 exercise_ref=exercise_ref,
                                                                                 survey_id=survey_id,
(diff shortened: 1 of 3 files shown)


swagger_server/controllers/survey_response.py
```diff
@@ -56,11 +56,21 @@ def add_survey_response(self, case_id, file):
...
+                ons_env.logger.debug('generating file name for upload with a survey ID of: {}'.format(survey_id))
+            else:
+                return self._invalid_upload()
+
+            # request survey service data from the API gateway
+            survey_service_data = self._get_survey_service(survey_id)
+
+            if survey_service_data:
+                survey_ref = survey_service_data('surveyRef')
+                ons_env.logger.debug('generating file name for survey reference id for upload with a survey ID of: {}'.format(survey_ref))
```
LJBabbage: I would expect this debug log message to be in the self._generate_file_name function

swagger_server/controllers/survey_response.py
```diff
@@ -127,6 +137,39 @@ def _get_collection_exercise(self, collection_exercise_id):
             ons_env.logger.debug('Collection Exercise not found')
         return collection_exercise
 
+
+    def _get_survey_service(self, survey_id):
```
LJBabbage: The function name is a bit misleading, we are getting the survey_data not the service

swagger_server/controllers/survey_response.py
```diff
@@ -127,6 +137,39 @@ def _get_collection_exercise(self, collection_exercise_id):
...
+        """
+        Used to get the SurveyRef number from the survey service. This is used to construct the file name of an
```
LJBabbage: The  comment should really be what the function is doing, not what the thing calling it will be using it for. So the comment should be get survey data from the survey service etc

swagger_server/controllers/survey_response.py
```diff
@@ -127,6 +137,39 @@ def _get_collection_exercise(self, collection_exercise_id):
...
+        uploaded survey.
+        See: https://github.com/ONSdigital/rm-survey-service/blob/master/API.md
+
+        :param survey_id: The survey_id UUID to search with
+            e.g. http://localhost:8080/surveys/cb0711c3-0ac8-41d3-ae0e-567e5ea1ef87
+
+        :return: JSON e.g.
+
+        {
```
LJBabbage: A bit too verbose for me

nherriot: please leave this in. it's a god send in years to come. please read: >>> import this

PR #47, opened 2017-09-15
moves jwt keys out of the [default] group. this means that when the s…

nherriot:
Nothing special, just making sure our jwt keys are not available in the default section. Hence the system fails if you don't set them via an environment variable - or run ONS_ENV=travis or ONS_ENV=dev

### DIFF
--- a/config.ini
+++ b/config.ini
@@ -72,9 +72,6 @@ authentication = false
 ;   Encryption settings, for production purposes all these settings need to be
 ;   provided via environment variables.
 ;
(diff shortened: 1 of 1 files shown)


config.ini
```diff
@@ -122,6 +127,10 @@ authentication = false
 protocol = https
 db_service = ras-postgres
 db_schema = ras_ci
+crypto_key = ONS_DUMMY_KEY
+jwt_algorithm = HS256jwt
```
andrewmil: is this supposed to have `jwt` in it?

PR #48, opened 2017-10-02
Dockerise application

jwgwalton:


config.ini
```diff
@@ -29,15 +29,19 @@
 enable_database = yes
 drop_database = no
 ;
+;  PORT FROM CRIB SHEET
```
LJBabbage: What does port from crib sheet mean?

jwgwalton: I'm following the port mappings from marks developer environment, it shouldn't be there as a comment. I'll remove it

PR #49, opened 2017-10-03
adding basic auth to import script

andrewmil:
Updating import script
- updating urls to not use api-gateway
- adding basic auth to post request


LJBabbage:
I hovered this up in my refactor, so I am not sure how we want to proceed

andrewmil:
If it's been done elsewhere then we can close this, I just did this quick a few weeks ago off the back of something mark said

PR #51, opened 2017-10-23
Add manifest for cloudfoundry deployments

benjefferies:
Add manifest file for cloudfoundry deployments

1. Are the defaults set to sensible values?
1. Are we missing any manifest properties that should be there?
1. Could we strip them down further?


benjefferies:
Merged although code coverage was failing because this was not the cause

PR #50, opened 2017-10-16
Tbl102 refactor collection instrument

LJBabbage:
This is a complete refactor of ras-collection instrument. 

**Out of scope**
1. How files are transfered from and to the application
2. Name of end points and what is passed
3. Model structure

**Summary of changes**
1. Remove Swagger
2. Remove RAS common
3. Change config to yml 
4. Tidy up requirements
5. Remove un-used end points, there now should be only 4 for collection instrument and 1 for survey response
6. Increase test coverage above (95%)
7. Fix and enforce flake8
8. Improve logging and documentation
9. Update readme
10. Fix Codacy issues
11. General refactoring, renaming, function size etc

PR #52, opened 2017-10-30
Move response status to views

LJBabbage:
This branch is based of https://github.com/ONSdigital/ras-collection-instrument/pull/50, which needs to be merged first. It moves the response status logic to the views rather then being in the controller


application/controllers/survey_response.py
```diff
@@ -6,82 +6,37 @@
...
     The survey response from a respondent
     """
-    def add_survey_response(self, case_id, file):
+    def add_survey_response(self, case_id, file, file_name):
         """
         Encrypt and upload survey response to rabbitmq
         :param case_id: A case id
         :param file: A file object from which we can read the file contents
+        :param file_name: The filename
         :return: Returns status code and message
```
benjefferies: Doc needs updating

application/controllers/survey_response.py
```diff
@@ -6,82 +6,37 @@
...
+FILE_EXTENSION_ERROR = 'The spreadsheet must be in .xls or .xlsx format'
 FILE_NAME_LENGTH_ERROR = 'The file name of your spreadsheet must be less than 50 characters long'
-UPLOAD_SUCCESSFUL = 'Upload successful'
-UPLOAD_UNSUCCESSFUL = 'Upload failed'
-INVALID_UPLOAD = 'The upload must have valid case_id and a file attached'
-NOT_FOUND_SERVICE_RESPONSE = 'A request to an external service returned no data'
 QUEUE_NAME = 'Seft.Responses'
 
 
 class SurveyResponse(object):
```
benjefferies: This probably doesn't need to be a class and could just be a few functions

tests/controllers/test_collection_instrument.py
```diff
@@ -64,25 +61,14 @@ def test_get_instrument_by_search_string_invalid_search_value(self):
         with self.assertRaises(RasDatabaseError):
             self.collection_instrument.get_instrument_by_search_string('{\"COLLECTION_EXERCISE\": \"invalid_uuid\"}')
 
-    def test_get_instrument_by_id(self):
```
benjefferies: Do we need a test for get_instrument_json_by_id?

LJBabbage: The view test covers it

application/controllers/survey_response.py
LJBabbage: Yep, but maybe one for the future. Just need to get this PR in before it disintegrates. Its been alive far to long

PR #54, opened 2017-11-06
Removing tox

benjefferies:
Tox isn't being used by any build process and is currently failing in
it's current form. With pipenv being introduced the value of fixing the
tox configuration vs what it will give us means it's worth removing.


benjefferies:
Good point I'll close in that case

PR #57, opened 2017-11-29
Add ras common utils hash & version

gemmai95:


JamesGardiner:
@gemmai95 - the build for this failed due to an out of date Pipfile.lock...

PR #59, opened 2017-11-30
Re-dockerised ras-collection-instrument service

madeye-matt:
Can someone take a careful look at the changes in Pipfile.lock - they came about when I did 'pipenv install xlsxwriter' but they don't look right to me ...

### DIFF
--- a/config/config-docker.yaml
+++ b/config/config-docker.yaml
@@ -1,0 +1,31 @@
+service:
+    NAME: ras-collection-instrument
+    VERSION: 1.0.2
+    SCHEME: http
+    HOST: 0.0.0.0
+    PORT: 8082
+    DEBUG: False
+    LOG_LEVEL: info
+    ONS_CRYPTOKEY: crypto_key
+    SECURITY_USER_NAME: test_user_name
+    SECURITY_USER_PASSWORD : test_user_password
+    RABBITMQ_AMQP: rabbitmq_amqp
+    UPLOAD_FILE_EXTENSIONS: xls,xlsx
+    MAX_UPLOAD_FILE_NAME_LENGTH: 50
+
+dependencies:
+    ras-ci-db:
+        schema: ras_ci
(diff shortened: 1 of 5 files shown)


Pipfile
```diff
@@ -27,6 +27,7 @@ pika = "==0.10.0"
 cryptography = "==1.9"
 flask-httpauth = "==3.2.3"
 six = "==1.10.0"
+xlsxwriter = "*"
```
andrewmil: We should probably pin this to a specific version

config/config-docker.yaml
```diff
@@ -0,0 +1,31 @@
+service:
```
benjefferies: Do we need the config for docker or will overriding the environmental variables be good enough?

JamesGardiner: Unfortunately this service is not 12FA and doesn't get configured via it's environment. Instead a YAML file is used to configure the service as per https://github.com/ONSdigital/ras-collection-instrument/blob/20cce0c57516e373adf4d3b67321f40e696a717a/app.py#L14

JamesGardiner: I will raise a tech debt issue here and on trello to get this app configured via it's environment.

benjefferies: I think it has some support around env variables but don't know how it works fully https://github.com/ONSdigital/ras-common-utils/blob/master/ras_common_utils/ras_config/ras_config.py#L38

Pipfile
LJBabbage: It should be 1.0.2

config/config-docker.yaml
madeye-matt: I was unable to configure this application from the environment so what I did was to change run.py to allow an env variable to specify the config file so you could override the config from the environment.  I ran this approach past a couple of people and they were happy it was an improvment on what went before :-) I can't see what code changes are required here (short of rewriting the whole configuration piece, which is beyond the scope of this card)

benjefferies:
@madeye-matt pipfile.lock file looks out of date

PR #61, opened 2017-12-04
Add missing end points and tests

LJBabbage:
Two ends were overlooked on the refactor /download/<instrument_id> and /instrument_size/<instrument_id>. This PR reintroduces them

### DIFF
--- a/application/controllers/collection_instrument.py
+++ b/application/controllers/collection_instrument.py
@@ -143,10 +143,7 @@ class CollectionInstrument(object):
         """
 
         instrument = CollectionInstrument.get_instrument_by_id(instrument_id, session)
-
-        instrument_json = None
-        if instrument:
-            instrument_json = instrument.json
+        instrument_json = instrument.json if instrument else None
         return instrument_json
 
     @staticmethod
--- a/application/controllers/collection_instrument.py
+++ b/application/controllers/collection_instrument.py
@@ -9,7 +9,6 @@ from json import loads
 from ras_common_utils.ras_error.ras_error import RasError
 from structlog import get_logger
 
-
 log = get_logger()
 
 UPLOAD_SUCCESSFUL = 'The upload was successful'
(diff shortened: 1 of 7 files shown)


andrewmil:
Looks like we also need to make changes in frontstage-(api) to work with these endpoints, It looks like upload doesn't work either atm

application/controllers/collection_instrument.py
```diff
@@ -136,24 +135,51 @@ def get_instruments_by_exercise_id_csv(exercise_id, session=None):
...
         :return: formatted JSON version of the instrument
         """
 
-        log.info('Searching for instrument using id {}'.format(instrument_id))
+        instrument = CollectionInstrument.get_instrument_by_id(instrument_id, session)
+
         instrument_json = None
-        validate_uuid([instrument_id])
+        if instrument:
+            instrument_json = instrument.json
```
andrewmil: These 3 line can be replaced with 
`instrument_json = instrument.json if instrument else None`

application/views/collection_instrument_view.py
```diff
@@ -46,9 +46,34 @@ def collection_instrument_by_search_string():
...
+
+
+@auth.login_required
+@collection_instrument_view.route('/instrument_size/<instrument_id>', methods=['GET'])
+def instrument_size(instrument_id):
+    instrument = CollectionInstrument().get_instrument_json(instrument_id)
+
+    if instrument and 'len' in instrument:
+        return make_response(jsonify(instrument['len']), 200)
+    else:
```
andrewmil: Since we return in the if statement the else here is unnecessary. 
It might be cleaner to do 
```
if not instrument...:
    return (..., 404)
return make_response(jsonify(instrument['len']), 200)
```

application/controllers/collection_instrument.py
LJBabbage: Never a massive fan of the readability of this approach

application/views/collection_instrument_view.py
LJBabbage: Fair comment, will update

benjefferies:
@pricem14pc your right the with_db_session decorator does pass the session in however I think the with_db_session is unnecessary code and we could look at replacing it with http://flask-sqlalchemy.pocoo.org/2.3/ which would handle the database sessions around the http requests

PR #62, opened 2017-12-11
Return instrument size as string without \n

andrewmil:
Small fix to stop collection instrument size endpoint returning `\n` at the end of responses.

We were jsonifying the string when we didn't need to


andrewmil:
[Collection Instrument API returns \n when requesting file size](https://trello.com/c/FzGZ3g4o/25-collection-instrument-api-returns-n-when-requesting-file-size)

PR #63, opened 2017-12-12
Feature/add survey id to jwt for sdx

AndrewTorrance:
This is to pass what the survey ref (known as survey id by sdx) from the collection instrument survey to sdx so that sdx can identify the survey and hence be used to trigger storage of the information in survey specific locations downstream . 

In doing this work a bug was uncovered on master whereby survey_id (uuid) was being used to format the filename when survey_ref should have been used . This was a regression , this was also corrected as part of this work .

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,10 +11,13 @@ The API is specified [here](./API.md).
 This requires pipenv and tox to be installed:
 ```
 pip install pipenv
-pip install tox
 ```
 
 ## Tests
+Tests require tox to be installed.
+```
+pip install tox
+```
 To run the tests with [tox], install tox, then simply run the command `pipenv run tox` in the root of the project.
 tox will create a unique virtualenv, run the unit tests with py.test, then run flake8 coverage.
 
--- a/README.md
+++ b/README.md
@@ -7,6 +7,12 @@ This is the RAS Collection Instrument micro-service, responsible for the uploadi
 
 The API is specified [here](./API.md).
 
+## Environment
+This requires pipenv and tox to be installed:
+```
+pip install pipenv
+pip install tox
+```
 
 ## Tests
 To run the tests with [tox], install tox, then simply run the command `pipenv run tox` in the root of the project.
--- a/README.md
+++ b/README.md
@@ -8,18 +8,19 @@ This is the RAS Collection Instrument micro-service, responsible for the uploadi
 The API is specified [here](./API.md).
 
 ## Environment
-This requires pipenv and tox to be installed:
+This requires pipenv to be installed:
 ```
 pip install pipenv
 ```
 
 ## Tests
-Tests require tox to be installed.
+To run the tests, install the dev dependencies using Pipenv and run tox within the virtual environment, 
+which will run the unit tests with py.test, then run flake8 coverage.
 ```
-pip install tox
+pipenv install --dev
+pipenv run tox
 ```
-To run the tests with [tox], install tox, then simply run the command `pipenv run tox` in the root of the project.
-tox will create a unique virtualenv, run the unit tests with py.test, then run flake8 coverage.
+
 
 To install and run:
 ``` bash
--- a/application/controllers/survey_response.py
+++ b/application/controllers/survey_response.py
@@ -157,12 +157,12 @@ class SurveyResponse(object):
         check_letter = 'X'  # Business related file name requirement
         time_date_stamp = time.strftime("%Y%m%d%H%M%S")
         file_name = "{ru}{check_letter}_{exercise_ref}_" \
-                    "{survey_id}_{time_date_stamp}{file_format}".format(ru=ru,
-                                                                        check_letter=check_letter,
-                                                                        exercise_ref=exercise_ref,
-                                                                        survey_id=survey_ref,
-                                                                        time_date_stamp=time_date_stamp,
-                                                                        file_format=file_extension)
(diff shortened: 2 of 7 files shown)


AndrewTorrance:
[Update RAS collection instrument to add survey id to JWT (5)](https://trello.com/c/pJCYcsHM/561-update-ras-collection-instrument-to-add-survey-id-to-jwt-5)

AndrewTorrance:
[Update RAS collection instrument to add survey id to JWT (5)](https://trello.com/c/pJCYcsHM/561-update-ras-collection-instrument-to-add-survey-id-to-jwt-5)

AndrewTorrance:
[Update RAS collection instrument to add survey id to JWT (5)](https://trello.com/c/pJCYcsHM/561-update-ras-collection-instrument-to-add-survey-id-to-jwt-5)

application/controllers/survey_response.py
```diff
@@ -112,32 +121,35 @@ def _encrypt_message(message_json):
         encrypter = Encrypter(json_secret_keys)
         return encrypter.encrypt(message_json)
 
-    def generate_file_name(self, case_id, file_extension):
+    def get_file_name_and_survey_ref(self, case_id, file_extension):
```
andrewmil: I think combining these two bit of functionality in one function here isn't ideal, returning 2 values will making testing a bit more convoluted.
The actual creation of the file name should probably be separate from the calls to get the info for it.
This is probably more to do with the current structure of the code rather than this PR

README.md
```diff
@@ -7,6 +7,12 @@ This is the RAS Collection Instrument micro-service, responsible for the uploadi
 
 The API is specified [here](./API.md).
 
+## Environment
+This requires pipenv and tox to be installed:
+```
+pip install pipenv
+pip install tox
```
andrewmil: I think tox is only required for testing so should maybe be in that section?

application/views/survey_responses_view.py
```diff
@@ -29,12 +29,14 @@ def add_survey_response(case_id):
...
-        file_name = survey_response.generate_file_name(case_id, file_extension)
-        print(file_name)
+        file_name, survey_ref = survey_response.get_file_name_and_survey_ref(case_id, file_extension)
+
         if not file_name:
             return make_response(MISSING_DATA, 404)
 
-        upload_success = survey_response.add_survey_response(case_id, file, file_name)
+        upload_success = survey_response.add_survey_response(case_id, file, file_name, survey_ref)
+        log.info("survey_responses/{0} ")
```
andrewmil: Does the `{0}` do anything here?

AndrewTorrance: Your right , shouldn't be like that , the log line shouldn't be there at all. The add_survey_response logs , I started doing this with a view to adding logging , changed my mind and forgot to delete the code . Thanks , good spot

application/controllers/survey_response.py
AndrewTorrance: You are right , it rubbed me up the wrong way . But when you look at the code it goes to case_group, collection_exercise and then survey which is 3 service calls . And its difficult to extract either a survey_ref or filename without doing the same expensive steps . So I added a comment as to why I did this, personally in this case I would leave as is . If I knew that the services all used some sort of caching then I would say split them . This might be an argument for adding caching to all services . It does not have to be complex , it could be a simple dictionary with a request lifespan . If we had that then calling these services for the same information would involve 1 http call and one dictionary lookup. Then I would agree , split them . It might be worth raising that . Its a trivial thing to add and could lead to better looking code ? I might pick that up with Mark , or feel free to raise it ?

README.md
AndrewTorrance: Good call, changing

README.md
```diff
@@ -7,8 +7,17 @@ This is the RAS Collection Instrument micro-service, responsible for the uploadi
 
 The API is specified [here](./API.md).
 
+## Environment
+This requires pipenv and tox to be installed:
```
jcox-dev: -and tox

README.md
```diff
@@ -7,8 +7,17 @@ This is the RAS Collection Instrument micro-service, responsible for the uploadi
...
+```
+pip install pipenv
+```
 
 ## Tests
+Tests require tox to be installed.
+```
+pip install tox
```
jcox-dev: Tox is already listed in the Pipfile's dev dependencies. So perhaps line 17-19 are not needed here and line 21-22 can be updated to something like:

"To run the tests, install the dev dependencies using Pipenv and run tox within the virtual environment, which will run the unit tests with py.test, then run flake8 coverage.
```
pipenv install --dev
pipenv run tox
```
"

AndrewTorrance: Good call. The confusion arose on my part because I ran the tests before pip install. Thanks for the clarification

application/controllers/survey_response.py
andrewmil: Functionally it all seems to be working fine so I'm happy to leave as is.

I think a solution could be to make all the required external calls in the view sequentially then have a generate_file_name function which takes all the required elements (exercise_ref/survey_ref) directly

pricem14pc: It would be good to understand how these python 'classes' are handled re: scope. Having an equivalent of a request scoped spring 'bean' would probably give us some options. We will be looking at reddis for caching at some point, which I imagine will initially cache (the relatively static) survey and CE meta data that we are  concerned with here.

Issue #39, opened 2017-09-05
Rabbit Submitter is not using Publisher Confirms


JamesGardiner:
https://github.com/ONSdigital/ras-collection-instrument/blob/master/swagger_server/controllers/submitter/rabbitmq_submitter.py

The above module deals with publishing messages to a RabbitMQ cluster. Ideally, this should use the publisher confirms method of dealing with the broker, and that the AMQP RPC `immediate` bit is set to false and the `mandatory` bit is set to true.

When setup in this way, the AMQP server will return a `Basic.Ack` when a message is written to a master queue and all of its slave queues, allowing the publisher to retain a reference to the submitted data until it is confirmed as being on the queue.

JamesGardiner:
![](https://github.trello.services/images/mini-trello-icon.png) [Collection instrument does not use the publisher confirms model for safe message transport](https://trello.com/c/M6VlovaG/9-collection-instrument-does-not-use-the-publisher-confirms-model-for-safe-message-transport)

jcox-dev:
@JamesGardiner I expect we can close this now?

PR #64, opened 2017-12-13
Use sdc-rabbit QueuePublisher for survey responses

andrewmil:
The new imported QueuePublisher is now configured to use publisher confirms, by setting `mandatory=True` in the publish function.

Since we're now usin an imported package the file testing `RabbitMQSubmitter` can be removed

[issue](https://github.com/ONSdigital/ras-collection-instrument/issues/39)
[trello](https://trello.com/c/M6VlovaG/9-collection-instrument-does-not-use-the-publisher-confirms-model-for-safe-message-transport)


andrewmil:
[Collection instrument does not use the publisher confirms model for safe message transport](https://trello.com/c/M6VlovaG/9-collection-instrument-does-not-use-the-publisher-confirms-model-for-safe-message-transport)

application/controllers/survey_response.py
```diff
@@ -79,9 +83,15 @@ def _send_message_to_rabbitmq(encrypted_message, tx_id):
 
         log.info('Getting environmental details for rabbit')
         rabbitmq_amqp = current_app.config.get('RABBITMQ_AMQP')
-        rabbit_mq = RabbitMQSubmitter(rabbitmq_amqp)
+        publisher = QueuePublisher([rabbitmq_amqp], QUEUE_NAME)
+        try:
+            result = publisher.publish_message(encrypted_message, headers={"tx_id": tx_id},
+                                               immediate=False, mandatory=True)
+        except PublishMessageError:
+            log.error()
```
LJBabbage: Do we need this? The SDC import (QueuePublisher) does log the issue

application/controllers/survey_response.py
```diff
@@ -2,13 +2,17 @@
...
 from flask import current_app
+from sdc.rabbit.exceptions import PublishMessageError
+from sdc.rabbit.publisher import QueuePublisher
 from structlog import get_logger
-from application.controllers.service_helper import get_case_group, get_collection_exercise, get_survey_ref
+
+from application.controllers.helper import (is_valid_file_extension, is_valid_file_name_length,
+                                            convert_file_object_to_string_base64)
+from application.controllers.json_encrypter import Encrypter
+from application.controllers.service_helper import (get_case_group, get_collection_exercise,
```
LJBabbage: Can stick this on one line and keep flake8 happy would then not need the brackets

LJBabbage:
This is fine to merge, bar the odd comment. Its not going to cause any issues. Good work

application/controllers/survey_response.py
andrewmil: Do we have an agreed max line-length? I've been using 100 across frontstage/backstage(-api)

andrewmil: Do we set the flake8 settings anywhere in the repo? I'm confused how it passes when flake8's default max-line-length is ~80?

application/controllers/survey_response.py
```diff
@@ -79,9 +82,14 @@ def _send_message_to_rabbitmq(encrypted_message, tx_id):
...
-        rabbit_mq = RabbitMQSubmitter(rabbitmq_amqp)
+        publisher = QueuePublisher([rabbitmq_amqp], QUEUE_NAME)
+        try:
+            result = publisher.publish_message(encrypted_message, headers={"tx_id": tx_id},
+                                               immediate=False, mandatory=True)
+        except PublishMessageError:
+            return False
 
-        if rabbit_mq.send_message(encrypted_message, QUEUE_NAME, tx_id):
+        if result:
```
tej99: Is there a need for this check? If the message does not get published it throws PublishMessageError and returns  false anyway.

andrewmil: You're right. Just talked to James and when we have the `mandatory=True` flag on the publisher it throws an exception rather than returning False.

PR #80, opened 2018-01-26
Add resilience to database errors or non-existence

benjefferies:
Retry database connection on DatabaseError every 2s for up to 30s. This
allows the app to be more robust when starting up. For instance when
running up apps via docker-compose the app should be able to recover
if the database is not up first.


benjefferies:
Good point. I'll see how hard it would be

PR #81, opened 2018-01-26
update logging standards

rhiGriff:
Updating logging to conform with SDC standards

### DIFF
--- a/application/controllers/collection_instrument.py
+++ b/application/controllers/collection_instrument.py
@@ -72,7 +72,7 @@ class CollectionInstrument(object):
         :return 'UPLOAD_SUCCESSFUL' if the upload completed
         """
 
-        log.info('Upload exercise: ', exercise_id=exercise_id.format)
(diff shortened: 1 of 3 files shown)


application/controllers/collection_instrument.py
```diff
@@ -72,7 +72,7 @@ def upload_instrument(self, exercise_id, file, ru_ref=None, classifiers=None, se
         :return 'UPLOAD_SUCCESSFUL' if the upload completed
         """
 
-        log.info('Upload exercise: {}'.format(exercise_id))
+        log.info('Upload exercise: ', exercise_id=exercise_id.format)
```
andrewmil: The standard way we do logging in the other services would be 
`log.info('Upload exercise', exercise_id=exercise_id)`
Since everything gets put into json format the colon doesn't really fit here.
Also shouldn't have `.format` here

rhiGriff: I missed that .format good spot 👍

PR #82, opened 2018-01-26
Use pycryptodome instead of pycryto

benjefferies:
`pipenv check` spots a vulnerability defined here
https://www.cvedetails.com/cve/CVE-2013-7459/. A stackoverflow post
suggests pycryptodome is a drop in replacement for pycrypto
https://stackoverflow.com/questions/41813030/problems-with-installation-pycrypto-in-python-3-6

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -20,7 +20,6 @@ gunicorn = "==19.7.1"
 sqlalchemy-utils = "==0.32.14"
 sqlalchemy = "==1.1.10"
 "psycopg2" = "==2.7.1"
-pycrypto = "==2.6.1"
 structlog = "==17.2.0"
 werkzeug = "==0.12.2"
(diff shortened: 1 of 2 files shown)


Pipfile
```diff
@@ -33,6 +32,7 @@ xlsxwriter = "==1.0.2"
 retrying = "==1.3.3"
 sdc-rabbit = "==1.2.1"
 alembic = "==0.9.6"
+pycryptodome = "*"
```
andrewmil: Should this be bound to a specific version?

benjefferies: Personally I think all the pinning we do is pointless because the pipfile.lock does all the pinning but I am happy to pin it for consistency (https://github.com/pypa/pipenv/issues/954)

PR #83, opened 2018-01-29
Make the survey id from the search return as a array

LJBabbage:
This is a temporary fix for bricks and blocks. Although survey_id shouldn't be an array, this has been done to allow another service to work. A TBL will be raised to fix the other service properly


davidmort:
are there any tests that need updating for this change?

PR #92, opened 2018-02-13
US042 Add retry for rabbitmq initialisation

jcox-dev:
The rabbitmq instance may not yet be ready to accept connections when the collection instrument service is started. This retries the initialisation steps (and therefore the connection to the provided rabbitmq URL) repeatedly until it fails (similar to init for the database).

### DIFF
--- a/app.py
+++ b/app.py
@@ -25,6 +25,10 @@ except RetryError:
     logger.exception('Failed to initialise database')
     exit(1)
 
-initialise_rabbit(app)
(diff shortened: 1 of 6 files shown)


jcox-dev:
[US042: Create and View 'Ready for Review' State of a CE [M]](https://trello.com/c/Fb6CS3es/36-us042-create-and-view-ready-for-review-state-of-a-ce-m)

PR #95, opened 2018-02-22
Us029 get ci by type

tej99:
Add ability to search CIs by CI type column.
Also change logic of returning CI seft file details to return None if the CI is not associated with a SEFT file.


tej99:
[US029: INT Select/Load EQ Collection Instruments [XXL]](https://trello.com/c/QharabAR/42-us029-int-select-load-eq-collection-instruments-xxl)

PR #100, opened 2018-02-27
change FORM_TYPE to form_type

tej99:


tej99:
[US029: INT Select/Load EQ Collection Instruments [XXL]](https://trello.com/c/QharabAR/42-us029-int-select-load-eq-collection-instruments-xxl)

PR #104, opened 2018-03-14
Search FORM_TYPE classfier with lowercase

tej99:
Search string for classifier FORM_TYPE will be caps but the 'form_type' classifier is stored lowercase. This PR allows to search using upper case FORM_TYPE for classifier 'form_type'.


jwgwalton:
@tej99 can we get around this by ignoring case. I think this is a jsonb field so not sure if possible?

tej99:
@jwgwalton changed the logic so it just make the classifier lower case by default now.

tej99:
[TBL133: CI association to CI selection classifier (XL)](https://trello.com/c/iIFH4z7r/54-tbl133-ci-association-to-ci-selection-classifier-xl)

PR #112, opened 2018-05-03
Fallback to using cloudfoundry queue

benjefferies:
If a rabbit queue is defined for collection instrument use it otherwise
use the environmental variables. This allows us to use a rabbit mq in
cloudfoundry for testing purposes


application/cloud/cloudfoundry.py
```diff
@@ -13,3 +13,10 @@ def detected(self):
     @property
     def db(self):
         return self._cf_env.get_service(name='ras-ci-db')
+
+    @property
+    def queue(self):
+        try:
+            return self._cf_env.get_service(name='ras-rabbitmq')
+        except StopIteration:
```
tej99: Looks like the stop iteration is handled in get_service, the next default is set to None.

PR #115, opened 2018-06-07
Make db connection pool configurable

tej99:
To improve performance of the system the db connection pool needs to be increased to sustain the load. This PR allows us to configure the pool size with environment variables.

### DIFF
--- a/config.py
+++ b/config.py
@@ -25,6 +25,10 @@ class Config(object):
     RABBITMQ_AMQP_COLLECTION_INSTRUMENT = cf.rm_queue_uri
     RABBITMQ_AMQP_SURVEY_RESPONSE = cf.sdx_queue_uri
     DATABASE_URI = cf.db_uri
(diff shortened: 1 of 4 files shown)


tej99:
[Optimise DB Connection Settings](https://trello.com/c/SwXkyTRo/44-optimise-db-connection-settings)

config.py
```diff
@@ -78,3 +82,6 @@ class TestingConfig(Config):
     DATABASE_SCHEMA = 'ras_ci'
     ONS_CRYPTOKEY = 'somethingsecure'
     JSON_SECRET_KEYS = open("./tests/files/keys.json").read()
+    DB_POOL_SIZE = 5
+    DB_MAX_OVERFLOW = 10
+    DB_POOL_RECYCLE = -1
```
andrewmil: ^ These are not required as they'll be set using the default Config object

PR #121, opened 2018-07-06
Publish message after transaction

benjefferies:
# Motivation and Context
We were seeing intermittent issues where collection exercises would
fail to be set into `ready for live` state. We observed in the logs that
collection instrument counts from the collection exercise service was 0
after collection instrument had said it have saved a collection
instrument. This was likely because collection exercise processed the
message before the data changes were committed to the database. This change publishes the message after the data has been saved. The
downside to this is that if a message failed to publish then the data
would not be rolled back. To manage this scenario we log at error level
with relevant information

The theory of the issue
1. Upload collection instrument (CI service) - https://github.com/ONSdigital/ras-collection-instrument/blob/master/application/views/collection_instrument_view.py#L33
2. Save collection instrument to the database (uncommitted) (CI service) - https://github.com/ONSdigital/ras-collection-instrument/blob/master/application/controllers/collection_instrument.py#L97
3. Post message to rabbit for collection exercise to pick up (CI service) - https://github.com/ONSdigital/ras-collection-instrument/blob/master/application/controllers/collection_instrument.py#L98
4. Collection exercise consumes message (Collex service) - https://github.com/ONSdigital/rm-collection-exercise-service/blob/dcd68bf63ed0fa2f492f856b45cbd9e16c0025a0/src/main/java/uk/gov/ons/ctp/response/collection/exercise/message/impl/CollectionInstrumentLoadedReceiver.java#L19
5. Collection exercise calls collection instrument to count the CIs (Collex service) - https://github.com/ONSdigital/rm-collection-exercise-service/blob/10d13f7332eb77b5ac5ec368478fad949e97e1e6/src/main/java/uk/gov/ons/ctp/response/collection/exercise/service/impl/CollectionExerciseServiceImpl.java#L685
6. Commit data to database (CI service) - https://github.com/ONSdigital/ras-collection-instrument/blob/master/application/controllers/collection_instrument.py#L102

# What has changed
Publish message only after committing data

# How to test?
Not sure. Been testing in CI

### DIFF
--- a/application/controllers/collection_instrument.py
+++ b/application/controllers/collection_instrument.py
@@ -1,8 +1,9 @@
 import logging
-import structlog
 import uuid
-
 from json import dumps, loads
+
+import structlog
+
 from application.controllers.cryptographer import Cryptographer
 from application.controllers.helper import validate_uuid
 from application.controllers.rabbit_helper import initialise_rabbitmq_exchange, send_message_to_rabbitmq_exchange
@@ -95,8 +96,6 @@ class CollectionInstrument(object):
             instrument.classifiers = loads(classifiers)
 
         session.add(instrument)
-        if not self.publish_uploaded_collection_instrument(exercise_id, instrument.instrument_id):
-            raise RasError('Failed to publish upload message', 500)
         return instrument
 
     @with_db_session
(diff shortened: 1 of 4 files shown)


davidmort:
Your jedi debugging skills are strong @benjefferies  :)

PR #124, opened 2018-08-13
Initialise db and rabbit at app create time

jcox-dev:
# Motivation and Context

Setting up the database and rabbit is required as part of the CI app creation so this has been moved inside the `create_app` function. This also logs out which config object the setup has used and added a rabbit container to the compose file.

NB: found that the tests weren't patching the rabbit service correctly (resulting in false positives) so have updated accordingly.

# Test
`make test`


.travis.yml
```diff
@@ -4,9 +4,6 @@ services:
 language: python
 python:
   - "3.6"
-before_install:
-  - docker-compose --verbose build
-  - docker-compose up -d
```
jcox-dev: Now part of the `tox` script.

jcox-dev:
[Log out config used in Python services](https://trello.com/c/Tctoorqs/251-log-out-config-used-in-python-services)

jcox-dev:
[Log out the config used at CI startup](https://trello.com/c/232KnUQN/252-log-out-the-config-used-at-ci-startup)

emilioward:
Whats the best way to tests this?

jcox-dev:
@emilioward `make test` 👍

PR #127, opened 2018-08-22
Fix import ordering

jcox-dev:
Running PyCharm's optimiser ✨


jcox-dev:
@benjefferies fixed in Dave's PR.

jcox-dev:
[Cleanup order of imports in Python services](https://trello.com/c/ulBo8OUj/259-cleanup-order-of-imports-in-python-services)

PR #137, opened 2019-01-21
Update dockerfile to use gunicorn to run app

AdamHawtin:
# Motivation and Context
The current Dockerfile runs the app in the flask development server which is not production suitable. This changes the Dockerfile to use Gunicorn to run the app with a configurable number of workers.

# What has changed
- Use entrypoint script in dockerfile
- Make number of gunicorn workers configurable by environment variable

# How to test?
You can run this docker image locally with ras-rm-docker-dev or use the branch tag in docker hub to run it in the cloud. Note that I needed to increase the default rm-kubernetes memory limits to run more than 4 workers.

# Links
https://trello.com/c/1tmchm16/439-collection-instruments-need-to-run-in-gunicorn-in-docker

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,5 +6,4 @@ EXPOSE 8002
 RUN apt-get update -y && apt-get install -y python-pip && apt-get install -y curl
 RUN pip3 install pipenv==8.3.1 && pipenv install --deploy --system
 
-ENTRYPOINT ["python3"]
-CMD ["run.py"]
+ENTRYPOINT ["sh", "entrypoint.sh"]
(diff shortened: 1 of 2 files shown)


AdamHawtin:
[Collection instruments need to run in GUnicorn in Docker](https://trello.com/c/1tmchm16/439-collection-instruments-need-to-run-in-gunicorn-in-docker)

AdamHawtin:
@jcox-dev This merging into an rm-census branch, not master (I'll update the description!). However it does make sense to keep a development Dockerfile for easier debugging and if we separate the prod file we can keep merging to master for now.

jcox-dev:
@AdamHawtin I didn't notice the target branch - thanks for pointing that out. Perhaps we make the distinction between the dev/prod when it comes to merging to master as part of another ticket?

PR #155, opened 2020-02-21
Update helm charts and dockerfile for k8s

insacuri:
# Motivation and Context

# What has changed

# How to test?

# Links


_infra/helm/collection-instrument/templates/deployment.yaml
```diff
@@ -6,30 +6,30 @@ spec:
   replicas: 1
```
jterryons: this should be templated too

PR #184, opened 2020-06-23
adding 0 byte rejection functionality

bkhan2:
# Motivation and Context
to reject 0 byte files.

# What has changed
added a function and a test for that function.


application/controllers/survey_response.py
```diff
@@ -34,10 +34,16 @@ def add_survey_response(self, case_id, file, file_name, survey_ref):
...
         log.info('Adding survey response file', filename=file_name, case_id=case_id, survey_id=survey_ref, tx_id=tx_id)
-
         file_contents = file.read()
-        json_message = self._create_json_message_for_file(file_name, file_contents, case_id, survey_ref)
-        return send_message_to_rabbitmq_queue(json_message, tx_id, RABBIT_QUEUE_NAME)
+        file_size = len(file_contents)
+
+        if self.check_if_file_size_too_small(file_size):
+            log.info('File size is too small')
+            return 'False'
```
warrenbailey, line 42: I'm not sure about returning false as a string here, it should be the boolean type false.

I think it would be better to return a boolean and a string from this function. The boolean to indicate if it was successful and a string relating to the error message

application/views/survey_responses_view.py
```diff
@@ -44,10 +45,12 @@ def add_survey_response(case_id):
 
         upload_success = survey_response.add_survey_response(case_id, file, file_name, survey_ref)
 
+        if upload_success == "False":
+            return make_response(FILE_TOO_SMALL, 400)
+
```
warrenbailey, line 50: This feels a bit wrong as you're first checking that the upload_sucess variable does not match the string 'false' and then immediately afterwards it's checked as a boolean.

application/views/survey_responses_view.py
```diff
@@ -44,10 +45,12 @@ def add_survey_response(case_id):
...
         if upload_success:
```
warrenbailey, line 51: I'd be tempted to move the error variables into the survey response and then do something like

upload_success, error = survey_response.add_survey_response(case_id, file, file_name, survey_ref)

if upload_success:
  return make_response(UPLOAD_SUCCESSFUL, 200)
else:
     return make_response(error, 500)

application/controllers/survey_response.py
```diff
@@ -144,6 +150,12 @@ def get_file_name_and_survey_ref(self, case_id, file_extension):
 
         return file_name, survey_ref
 
+    @staticmethod
+    def check_if_file_size_too_small(upload_file):
+        if upload_file < 1:
```
jterryons, line 155:
```
return upload_file < 1
```

application/controllers/survey_response.py
```diff
@@ -34,10 +34,16 @@ def add_survey_response(self, case_id, file, file_name, survey_ref):
...
-
         file_contents = file.read()
-        json_message = self._create_json_message_for_file(file_name, file_contents, case_id, survey_ref)
-        return send_message_to_rabbitmq_queue(json_message, tx_id, RABBIT_QUEUE_NAME)
+        file_size = len(file_contents)
+
+        if self.check_if_file_size_too_small(file_size):
+            log.info('File size is too small')
+            return 'False'
+        elif not self.check_if_file_size_too_small(file_size):
```
jterryons, line 43: you have already performed this check above, no need to recheck here

PR #185, opened 2020-06-26
updated README.md and added swagger.yaml

anwilkie:
# Motivation and Context
Like the other services, the collection instrument service required swagger documentation.

# What has changed

- Added `swagger.yaml` file
- Added service information and improvement suggestions to `README.md`
- Removed `API.md`

# Links
[Trello card](https://trello.com/c/hg1Q7l06)


swagger.yaml
```diff
@@ -0,0 +1,493 @@
...
+          required: true
+          schema:
+            type: string
+            format: dictionary
+            example: '{"FORM_TYPE":%20"001"}'
+          description: Collection instrument classifiers
+      responses:
+        '200':
+          description: Successfully uploaded collection instrument
+          content: "The upload was successful"
```
jterryons, line 96: This is not quite right. Whats the content type? `application/json`? `text/plain`? If its JSON you need to supply a schema. its not enought to just have `content: "some string"`
look at the API.yaml in the other repos for examples.

swagger.yaml
```diff
@@ -0,0 +1,493 @@
...
+          schema:
+            type: string
+            example: 'file_name.xlsx'
+          description: Spreadsheet file
+        - in: query
+          name: classifiers
+          required: true
+          schema:
+            type: string
+            format: dictionary
```
jterryons, line 90: are we really inputting a raw JSON string as the `classifiers` query param?

swagger.yaml
```diff
@@ -0,0 +1,493 @@
...
+info:
+  title: collection instrument
+  description: A service for the uploading of collection instruments and exercises
+  version: "2.0"
+
+servers:
+  - url: https://localhost:8080/
+    description: The path used by the collection exercise service when port forwarded by kubectl (8080:8080)
+
+tags:
```
jterryons, line 11: this is a little tag heavy. Get rid of all tags other than info and collection-instrument for now.

swagger.yaml
anwilkie: I tried looking at the other API.yaml files, but I couldn't find any endpoints where only a message is returned. I have modified these `content:` tags to `text/plain` objects.

swagger.yaml
anwilkie: Sorry, this was an error caused by me sloppily copying and pasting stuff around. I've deleted this now.

swagger.yaml
anwilkie: I've reduced the tags to just `collection-instrument`, `info` and `survey-response`.

PR #187, opened 2020-07-10
Zero byte file rejection

bkhan2:
0 byte file rejection.
Added the funcitonality and better handling of the view required when a user uploads a 0 byte file.


application/controllers/survey_response.py
```diff
@@ -39,11 +47,15 @@ def add_survey_response(self, case_id, file, file_name, survey_ref):
...
-            log.info('File size is correct')
+            raise FileTooSmallError()
+        else:
             json_message = self._create_json_message_for_file(file_name, file_contents, case_id, survey_ref)
-            return send_message_to_rabbitmq_queue(json_message, tx_id, RABBIT_QUEUE_NAME)
+            sent = send_message_to_rabbitmq_queue(json_message, tx_id, RABBIT_QUEUE_NAME)
+            if sent:
+                return sent
+            else:
+                raise SurveyResponseError
```
warrenbailey, line 57: Couple of minor things

1. no need to return sent, as you're not using the value in the calling method anymore
1. missing () on raise SurveyResponseError 
1. might be worth adding a log statement if it fails

Something like
```
 sent = send_message_to_rabbitmq_queue(json_message, tx_id, RABBIT_QUEUE_NAME)
 if not sent:
     log.error("Unable to send file to rabbit queue")
     raise SurveyResponseError()
```

application/views/survey_responses_view.py
```diff
@@ -43,14 +43,17 @@ def add_survey_response(case_id):
...
-
-        if upload_success:
-            return make_response(UPLOAD_SUCCESSFUL, 200)
-        else:
+        except SurveyResponseError:
+            log.error('Something went wrong')
             return make_response(UPLOAD_UNSUCCESSFUL, 500)
+        else:
+            log.info('File processed')
+            return make_response(UPLOAD_SUCCESSFUL, 200)
```
warrenbailey, line 57: I don't think you need this else, just move the logic into the try statement

application/views/survey_responses_view.py
```diff
@@ -43,14 +43,17 @@ def add_survey_response(case_id):
         if not file_name:
             return make_response(MISSING_DATA, 404)
 
-        upload_success = survey_response.add_survey_response(case_id, file, file_name, survey_ref)
-
-        if upload_success == "False":
+        try:
+            log.info('Processing file')
+            survey_response.add_survey_response(case_id, file, file_name, survey_ref)
```
warrenbailey, line 48: i.e. this should go here
```
log.info('File processed')
return make_response(UPLOAD_SUCCESSFUL, 200)
```

application/views/survey_responses_view.py
```diff
@@ -43,14 +43,17 @@ def add_survey_response(case_id):
...
+        except FileTooSmallError:
+            log.error('File too small')
             return make_response(FILE_TOO_SMALL, 400)
-
-        if upload_success:
-            return make_response(UPLOAD_SUCCESSFUL, 200)
-        else:
+        except SurveyResponseError:
+            log.error('Something went wrong')
```
warrenbailey, line 53: You probably want a more descriptive logging message.

Though if you add it in the above method, it's probably not needed here too.

application/views/survey_responses_view.py
```diff
@@ -43,14 +43,17 @@ def add_survey_response(case_id):
...
             return make_response(MISSING_DATA, 404)
 
-        upload_success = survey_response.add_survey_response(case_id, file, file_name, survey_ref)
-
-        if upload_success == "False":
+        try:
+            log.info('Processing file')
+            survey_response.add_survey_response(case_id, file, file_name, survey_ref)
+        except FileTooSmallError:
+            log.error('File too small')
```
warrenbailey, line 50: You're going to end up with two log statements exactly the same here, as you're logging this above. This could probably just be removed.

application/controllers/survey_response.py
```diff
@@ -39,11 +47,15 @@ def add_survey_response(self, case_id, file, file_name, survey_ref):
...
-            log.info('File size is correct')
+            raise FileTooSmallError()
+        else:
             json_message = self._create_json_message_for_file(file_name, file_contents, case_id, survey_ref)
-            return send_message_to_rabbitmq_queue(json_message, tx_id, RABBIT_QUEUE_NAME)
+            sent = send_message_to_rabbitmq_queue(json_message, tx_id, RABBIT_QUEUE_NAME)
+            if not sent:
+                log.error("Unable to send file to rabbit queue")
+                raise SurveyResponseError()
+            else:
```
warrenbailey, line 57: This return else and return statement can be removed as you're no longer using in the surveys_response_view.py file.

I know it's used in the test_add_survey_response_success test but the the assertTrue can be removed there and replaced with:

try:
     self.survey_response.add_survey_response(case_id, file, filename, '023')
except Exception:
     self.fail("Should not raise an exception")

tests/controllers/test_survey_response.py
```diff
@@ -48,8 +48,11 @@ def test_add_survey_response_success(self):
             with patch('pika.BlockingConnection'):
                 status = self.survey_response.add_survey_response(case_id, file, filename, '023')
 
-        # Then the file uploads successfully
-        self.assertTrue(status)
+            # Then the file uploads successfully
```
insacuri, line 51: This should be rewitten to something like 

```python
with patch('pika.BlockingConnection'):
     try:
                self.survey_response.add_survey_response(case_id, file, filename, '023')
    except:
                self.fail("Should not raise an exception")
```

tests/views/test_survey_response_view.py
```diff
@@ -46,14 +47,17 @@ def test_add_survey_response_success(self):
             # When that file is post to the survey response end point
             response = self.client.post(
                 '/survey_response-api/v1/survey_responses/{case_id}'.
```
insacuri, line 49: it's minor but rewrite this to `f'/survey_response-api/v1/survey_responses/{case_id}'` so we can remove the .format

application/controllers/survey_response.py
```diff
@@ -39,11 +47,13 @@ def add_survey_response(self, case_id, file, file_name, survey_ref):
...
-            return 'False'
-        elif not self.check_if_file_size_too_small(file_size):
-            log.info('File size is correct')
+            raise FileTooSmallError()
+        else:
             json_message = self._create_json_message_for_file(file_name, file_contents, case_id, survey_ref)
-            return send_message_to_rabbitmq_queue(json_message, tx_id, RABBIT_QUEUE_NAME)
+            sent = send_message_to_rabbitmq_queue(json_message, tx_id, RABBIT_QUEUE_NAME)
+            if not sent:
+                log.error("Unable to send file to rabbit queue")
```
insacuri, line 55: Is there any other metadata we can attach to the logger for this? tx_id, case_id, survey_id (for examples)?

warrenbailey:
LGTM :)

PR #212, opened 2021-02-16
Update packages

insacuri:
# Motivation and Context

# What has changed

# How to test?

# Links

### DIFF
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -18,11 +18,10 @@
     "default": {
         "alembic": {
             "hashes": [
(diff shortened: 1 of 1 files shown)


jdfox:
gevent was upgraded about a month ago from 20.x to 21.x (Pipfile.lock here uses latest 21.1.2):
- http://www.gevent.org/changelog.html
- https://github.com/gevent/gevent/issues

...but if you're confident that our tests cover this, that should be OK.

jdfox (approved):

PR #232, opened 2021-05-17
update env var, and new gnu class

mybigopenbox:
# What and why?
Replace encryption method to gnu and it it better
# How to test?
IN the env
# Trello
S33 GPG Encryption For SEFT responses


Dockerfile
```diff
@@ -1,12 +1,12 @@
...
-RUN apt-get update && apt-get install -y build-essential curl gunicorn
+RUN apt-get update && apt-get install -y build-essential curl gunicorn gnupg
 RUN pip install pipenv
 
 WORKDIR /app
 
 COPY . /app
 RUN pipenv install --deploy --system
-
+RUN gpg --help
```
insacuri, line 10: Why are we running help as part of the build?

_infra/helm/collection-instrument/values.yaml
```diff
@@ -68,4 +68,7 @@ gcp:
     prefix: ""
   topic: topic
 
-saveSeftInGcp: false
+saveSeftInGcp: true
```
insacuri, line 71: We want this false by default to protect us from any mistakes.

application/controllers/gnu_encryptor.py
```diff
@@ -0,0 +1,28 @@
...
+        Encrypts the payload using the recipient values
+
+        :param payload: the value to encrypt
+        :param recipient: who is it for
+        :return: string of encrypted data
+        """
+        payload = 'hello world'
+        enc_data = self.gpg.encrypt(json.dumps(payload).encode('utf-8'), recipient, always_trust=True)
+        if not enc_data.ok:
+            print('ok: ', enc_data.ok)
```
insacuri, line 22: These shouldn't be print statements.  You should use a logger for this.  Something like
```python
logger.error("Failed to encrypt with gpg", status=enc_data.status, error=enc_data.stderr, recipient=recipient)
```

application/controllers/rabbit_helper.py
```diff
@@ -19,9 +20,16 @@ def _encrypt_message(message_json):
     :return: jwe as String
     """
     log.info('Encrypting JSON message')
-    json_secret_keys = current_app.config['JSON_SECRET_KEYS']
-    encrypter = Encrypter(json_secret_keys)
-    return encrypter.encrypt(message_json)
+
+    if current_app.config['SAVE_SEFT_IN_GCP']:
```
insacuri, line 24: Why is this living in the rabbit helper? The communication over rabbit will never use GPG.

insacuri (changes requested):

_infra/helm/collection-instrument/Chart.yaml
```diff
@@ -18,4 +18,4 @@ version: 1.1.4
 
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application.
-appVersion: 2.0.57
+appVersion: 2.0.58
```
warrenbailey, line 21: This a fairly big change, so lets manually change this to 2.1.0

_infra/helm/collection-instrument/values.yaml
```diff
@@ -69,3 +69,6 @@ gcp:
   topic: topic
 
 saveSeftInGcp: false
+
```
warrenbailey, line 72: is there a dev key we can use as the default rather than preprod?

_infra/helm/collection-instrument/values.yaml
```diff
@@ -69,3 +69,6 @@ gcp:
...
+gnu_recipient: sdx_preprod@ons.gov.uk
```
warrenbailey, line 73: Rather an underscores, these should be in camelCase to match the other vars, i.e. gnuRecipient 

or actually even better you could do:
```
gnu:
  recipient: sdx_preprod@ons.gov.uk
  fingerprint: A8F49D6EE2DE17F03ACF11A9BF16B2EB4D779991
```

then in the values.yaml

```
 - name: ONS_GNU_FINGERPRINT            value: "{{ .Values.gnu.fingerprint }}"          
 - name: ONS_GNU_RECIPIENT            value: "{{ .Values.gnu.recipient }}"
```

application/controllers/gcp_survey_response.py
```diff
@@ -106,11 +110,14 @@ def put_file_into_gcp_bucket(self, message: dict):
...
-        encrypted_message = _encrypt_message(message)
+        gnugpg_secret_keys = current_app.config['ONS_GNU_PUBLIC_CRYPTOKEY']
+        ons_gnu_fingerprint = current_app.config['ONS_GNU_FINGERPRINT']
+        encrypter = GNUEncrypter(gnugpg_secret_keys)
+        encrypted_message = encrypter.encrypt(message, ons_gnu_fingerprint)
         blob.upload_from_string(encrypted_message)
         bound_log.info('Successfully put file in bucket', filename=filename)
 
-    def put_message_into_pubsub(self, payload: dict):
+    def put_message_into_pubsub(self, payload: dict, tx_id):
```
warrenbailey, line 120: this probably should have a type hint now as payload does

application/controllers/gnu_encryptor.py
```diff
@@ -0,0 +1,31 @@
...
+
+    def encrypt(self, payload, recipient):
+        """
+        Encrypts the payload using the recipient values
+
+        :param payload: the value to encrypt
+        :param recipient: who is it for
+        :return: string of encrypted data
+        """
+        enc_data = self.gpg.encrypt(json.dumps(payload).encode('utf-8'), recipient, always_trust=True)
```
warrenbailey, line 25: We may need to change this because we're not going to be converting the spreadsheet to a JSON object anymore. The idea will be to encrypt the file as is with GPG and send that directly to the bucket.

Therefore you may need to remove the call to json.dumps and change how this function is called

warrenbailey (changes requested):

_infra/helm/collection-instrument/Chart.yaml
mybigopenbox: ok

_infra/helm/collection-instrument/values.yaml
mybigopenbox: we can create one, we would need to store the dev private key and need an email address

_infra/helm/collection-instrument/Chart.yaml
mybigopenbox: agreed

warrenbailey (approved):
Looks good, can we remove the labels and edit the title if it's ready for merge?

insacuri (approved):

PR #237, opened 2021-06-14
Removing rabbit communication to collex with rest

creativeappdevamit:
# What and why?

# How to test?

# Trello
https://trello.com/c/ITyML5G9/732-replace-rabbit-with-rest-call-ras-collection-instrument-to-rm-collection-exercise-service


application/controllers/service_helper.py
```diff
@@ -112,3 +112,25 @@ def service_request(service, endpoint, search_value):
...
+
+    try:
+        collection_exercise_url = current_app.config['COLLECTION_EXERCISE_URL']
+        url = f'{collection_exercise_url}/collection-instrument/link'
+        log.info('Making request to collection exercise to acknowledge instrument load')
+    except KeyError:
+        raise RasError("collection exercise service not configured", 500)
+
+    response = requests.post(url, json=json_message, auth=auth)
+    response.raise_for_status()
```
warrenbailey, line 135: What happens with the exceptions that this may throw? 

If we don't need to deal with them, do we need to call raise for status here?

warrenbailey (changes requested):

application/controllers/service_helper.py
creativeappdevamit: I think yes, we want to record that the communication between two micro service was not successful isn't? It should be recorded both places.. collex and collex instrument?

warrenbailey (approved):

PR #238, opened 2021-06-15
remove json wrapper gnu seft - DRAFT

mybigopenbox:
# What and why?
remove json wrapper from message
# How to test?

# Trello
S33 GPG Encryption For SEFT responses

### DIFF
--- a/application/controllers/gcp_survey_response.py
+++ b/application/controllers/gcp_survey_response.py
@@ -103,6 +103,7 @@ class GcpSurveyResponse:
                 raise ValueError('Error with filename for bucket ')
         except ValueError as e:
             bound_log.info(e, filename=filename)
+            raise
 
         if self.storage_client is None:
             self.storage_client = storage.Client(project=self.seft_pubsub_project)
--- a/application/controllers/gcp_survey_response.py
+++ b/application/controllers/gcp_survey_response.py
@@ -84,14 +84,14 @@ class GcpSurveyResponse:
                 bound_log.exception("A non-timeout error was raised when publishing to pubsub", payload=payload)
                 raise SurveyResponseError()
 
-    def put_file_into_gcp_bucket(self, message: str, filename: str):
+    def put_file_into_gcp_bucket(self, file_contents, filename: str):
         """
-        Takes the message  and puts it into a GCP bucket in encrypted form to be later used by SDX.
+        Takes the file_contents  and puts it into a GCP bucket in encrypted form to be later used by SDX.
 
         Note: The payload will almost certainly change once the encryption method between us and SDX is decided, but
         for now we'll put the same payload we were using in rabbit as a starting point.
 
-        :param message: contents of the collection instrument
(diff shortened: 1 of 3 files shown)


application/controllers/gcp_survey_response.py
```diff
@@ -113,7 +113,7 @@ def put_file_into_gcp_bucket(self, message: dict):
         gnugpg_secret_keys = current_app.config['ONS_GNU_PUBLIC_CRYPTOKEY']
         ons_gnu_fingerprint = current_app.config['ONS_GNU_FINGERPRINT']
         encrypter = GNUEncrypter(gnugpg_secret_keys)
-        encrypted_message = encrypter.encrypt(message, ons_gnu_fingerprint)
+        encrypted_message = encrypter.encrypt(message['file'], ons_gnu_fingerprint)
```
warrenbailey, line 116: I don't think this will work because the file attribute of message is a base64 encoded representation of the file contents where we need the actual bytes

warrenbailey (changes requested):

application/controllers/gcp_survey_response.py
```diff
@@ -86,36 +84,48 @@ def add_survey_response(self, case_id: str, file_contents, file_name: str, surve
                 bound_log.exception("A non-timeout error was raised when publishing to pubsub", payload=payload)
                 raise SurveyResponseError()
 
-    def put_file_into_gcp_bucket(self, message: dict):
+    def put_file_into_gcp_bucket(self, message: str, filename: str):
```
warrenbailey, line 87: is the str type hint correct here? I'd expect it to be a bytes array?

also I'm wondering if message is the correct parameter name, it might be better as file_contents?

application/controllers/gcp_survey_response.py
```diff
@@ -86,36 +84,48 @@ def add_survey_response(self, case_id: str, file_contents, file_name: str, surve
...
-            bound_log.info('Missing filename from the message', message=message)
-            raise
+        if self.seft_bucket_file_prefix:
+            filename = f"{self.seft_bucket_file_prefix}/{filename}"
+        blob = bucket.blob(filename)
         gnugpg_secret_keys = current_app.config['ONS_GNU_PUBLIC_CRYPTOKEY']
         ons_gnu_fingerprint = current_app.config['ONS_GNU_FINGERPRINT']
         encrypter = GNUEncrypter(gnugpg_secret_keys)
         encrypted_message = encrypter.encrypt(message, ons_gnu_fingerprint)
+        md5Hash = hashlib.md5(json.dumps(encrypted_message, sort_keys=True).encode()).hexdigest()
```
warrenbailey, line 119: I wouldn't of expected a json.dumps() here? 

Could it be

```
md5Hash = hashlib.md5(encrypted_message).hexdigest()

``

warrenbailey (changes requested):
This is looking great, a couple of minor things :)

warrenbailey (approved):

PR #239, opened 2021-06-17
Gnu seft

mybigopenbox:
# What and why?
change size of encrypted msg
# How to test?

# Trello


application/controllers/gcp_survey_response.py
```diff
@@ -117,7 +116,7 @@ def put_file_into_gcp_bucket(self, file_contents, filename: str):
         encrypter = GNUEncrypter(gnugpg_secret_keys)
         encrypted_message = encrypter.encrypt(file_contents, ons_gnu_fingerprint)
         md5sum = hashlib.md5(str(encrypted_message).encode()).hexdigest()
-        sizeInBytes = sys.getsizeof(encrypted_message)
+        sizeInBytes = len(str(encrypted_message))
```
warrenbailey, line 119: why does it need to be converted to a string first?

mybigopenbox: previously it complained about TypeError: object of type 'Crypt' has no len()

warrenbailey (approved):

PR #241, opened 2021-06-21
Gnu seft - add .gpg extention

mybigopenbox:
# What and why?
require ext to track down the right file
# How to test?

# Trello
https://trello.com/c/Yuj6r1GI/651-s33-gpg-encryption-for-seft-responses


mybigopenbox:
created new br

PR #245, opened 2021-06-30
Fix deprecation warning

insacuri:
# What and why?

When running the tests, we were getting warnings that `_app_ctx_stack.__ident_func__()` is going to be deprecated in werkzeug 2.1.  This PR fixes that and a few other things.

We were able to simply remove that as the guide in flask about this (found https://flask.palletsprojects.com/en/2.0.x/patterns/sqlalchemy/) doesn't use it.  When you don't provide a scopedfunc, it defaults to managing sessions on a per thread basis.  We run these apps in preprod/prod using gunicorn which (I believe) runs app multiple times over multiple threads, meaning it should be safe. 

One thing I added, for safety more than anything, was an explicit session removal at the end of each flask request, rather than relying on the session decorator to do it for us.  It's possible for someone to write a db query without  using the session decorator, and if that happens then it's possible that removing the session could be forgotten about.  Failing to this could cause memory leaks, connection pools getting full and other pain in the butt to diagnose issues!  In testing it myself I didn't see any downside to doing 2 `session.remove()` so I believe doing it twice (once in the session decorator and again on the request teardown) should be fine.

More reading around this can be found here: https://docs.sqlalchemy.org/en/14/orm/contextual.html#using-thread-local-scope-with-web-applications

This PR will also add isort. The changes will look like a lot but it's only import order so it shouldn't be a big deal!

# How to test?

# Trello


insacuri:
Doing isort and this fix at the same time was dumb.  Need to do this in 2 separate PRs

PR #251, opened 2021-09-01
initial commit

anwilkie:
# What and why?

The SEFT CIs needed to be added to a GCP bucket.

# How to test?

TBD

# Trello

[Card](https://trello.com/c/7t6DzDNz)

### DIFF
--- a/_infra/helm/collection-instrument/templates/deployment.yaml
+++ b/_infra/helm/collection-instrument/templates/deployment.yaml
@@ -275,5 +275,9 @@ spec:
             value: "{{ .Values.gcp.seftProject }}"
           - name: SEFT_PUBSUB_TOPIC
             value: "{{ .Values.gcp.seftTopic }}"
+          - name: SEFT_CI_BUCKET_NAME
+            value: "{{ .Values.seftCIBucket.name }}"
+          - name: SEFT_CI_BUCKET_FILE_PREFIX
+            value: "{{ .Values.seftCIBucket.prefix }}"
           resources:
             {{ toYaml .Values.resources.application | nindent 12 }}
--- a/_infra/helm/collection-instrument/values.yaml
+++ b/_infra/helm/collection-instrument/values.yaml
@@ -68,6 +68,10 @@ gcp:
     prefix: ""
   topic: topic
 
+seftCIBucket:
+  name: ras-rm-seft-ci-bucket-dev
+  prefix: ""
+
 saveSeftInGcp: false
 
 gnu:
--- a/application/controllers/collection_instrument.py
+++ b/application/controllers/collection_instrument.py
@@ -2,8 +2,11 @@ import logging
 from json import loads
 
 import structlog
+from flask import current_app
+from google.cloud import storage
 
 from application.controllers.cryptographer import Cryptographer
+from application.controllers.gnu_encryptor import GNUEncrypter
 from application.controllers.helper import validate_uuid
 from application.controllers.service_helper import (
     collection_instrument_link,
@@ -107,8 +110,31 @@ class CollectionInstrument(object):
         if classifiers:
             instrument.classifiers = loads(classifiers)
 
+        try:
+            self.send_instrument_to_bucket(instrument)
+        except Exception:
+            log.error("An error occurred when trying to put SEFT collection instrument in bucket",
+                      instrument=instrument)
+
         session.add(instrument)
         return instrument
+
+    def send_instrument_to_bucket(self, instrument):
+        storage_client = storage.Client()
+
+        bucket_name = current_app.config.get("SEFT_CI_BUCKET_NAME")
+        blob_name = "SEFT CIs"
+
+        bucket = storage_client.bucket(bucket_name)
+        blob = bucket.blob(blob_name)
+        gnugpg_secret_keys = current_app.config["ONS_GNU_PUBLIC_CRYPTOKEY"]
+        ons_gnu_fingerprint = current_app.config["ONS_GNU_FINGERPRINT"]
+        encrypter = GNUEncrypter(gnugpg_secret_keys)
+        encrypted_instrument = encrypter.encrypt(instrument, ons_gnu_fingerprint)
+        blob.upload_from_string(encrypted_instrument)
+        log.info("Successfully put SEFT collection instrument in bucket")
+
+        return
 
     @with_db_session
     def patch_seft_instrument(self, instrument_id: str, file, session):
--- a/config.py
+++ b/config.py
@@ -29,6 +29,8 @@ class Config(object):
     SEFT_BUCKET_FILE_PREFIX = os.getenv("SEFT_BUCKET_FILE_PREFIX")
     SEFT_PUBSUB_PROJECT = os.getenv("SEFT_PUBSUB_PROJECT", "ras-rm-sandbox")
     SEFT_PUBSUB_TOPIC = os.getenv("SEFT_PUBSUB_TOPIC", "ras-rm-notify-test")
+    SEFT_CI_BUCKET_NAME = os.getenv("SEFT_CI_BUCKET_NAME")
+    SEFT_CI_BUCKET_FILE_PREFIX = os.getenv("SEFT_CI_BUCKET_FILE_PREFIX")
 
     UPLOAD_FILE_EXTENSIONS = "xls,xlsx"
 


_infra/helm/collection-instrument/values.yaml
```diff
@@ -58,6 +58,14 @@ gcp:
     prefix: ""
   topic: topic
 
+gcpDev:
+  project: ras-rm-dev
+  bucket:
+    name: "ras-rm-seft-ci-dev"
+    prefix: ""
+
+saveSeftInGcp: false
```
insacuri, line 67: This got removed in another PR.  This needs to be removed now

_infra/helm/collection-instrument/values.yaml
```diff
@@ -58,6 +58,14 @@ gcp:
     prefix: ""
   topic: topic
 
+gcpDev:
```
insacuri, line 61: This name is pretty bad.  Also the config is laid out wrong.  You can remove the `project` as it's the same between them.
The `bucket` should be renamed to `collectionInstrumentBucket` or something like it, should be in the `gcp` block above

_infra/helm/collection-instrument/templates/deployment.yaml
```diff
@@ -195,5 +195,11 @@ spec:
             value: "{{ .Values.gcp.seftProject }}"
           - name: SEFT_PUBSUB_TOPIC
             value: "{{ .Values.gcp.seftTopic }}"
+          - name: SEFT_CI_BUCKET_NAME
+            value: "{{ .Values.gcpDev.bucket.name }}"
+          - name: GOOGLE_CLOUD_PROJECT
+            value: "{{ .Values.gcpDev.project }}"
+          - name: SEFT_CI_BUCKET_FILE_PREFIX
+            value: "{{ .Values.gcpDev.bucket.prefix }}"
```
insacuri, line 203: These value locations will need to change based on my other comments

_infra/helm/collection-instrument/templates/deployment.yaml
```diff
@@ -195,5 +195,11 @@ spec:
             value: "{{ .Values.gcp.seftProject }}"
           - name: SEFT_PUBSUB_TOPIC
             value: "{{ .Values.gcp.seftTopic }}"
+          - name: SEFT_CI_BUCKET_NAME
+            value: "{{ .Values.gcpDev.bucket.name }}"
+          - name: GOOGLE_CLOUD_PROJECT
```
insacuri, line 200: The project is defined twice. We should only have 1 item and have both uses of it use the same environment variable.

config.py
```diff
@@ -24,6 +24,9 @@ class Config(object):
     SEFT_BUCKET_FILE_PREFIX = os.getenv("SEFT_BUCKET_FILE_PREFIX")
     SEFT_PUBSUB_PROJECT = os.getenv("SEFT_PUBSUB_PROJECT", "ras-rm-sandbox")
     SEFT_PUBSUB_TOPIC = os.getenv("SEFT_PUBSUB_TOPIC", "ras-rm-notify-test")
+    SEFT_CI_BUCKET_NAME = os.getenv("SEFT_CI_BUCKET_NAME")
+    GOOGLE_CLOUD_PROJECT = os.getenv("GOOGLE_CLOUD_PROJECT")
```
insacuri, line 28: This field might be okay to keep.  Can you change the bit that uses `SEFT_PUBSUB_PROJECT` to use this instead and remove the old value

migrations/versions/d975ea83b712_add_seft_ci_file_location.py
```diff
@@ -0,0 +1,25 @@
...
+depends_on = None
+
+
+def upgrade():
+    op.add_column("instrument", sa.Column("file_location", String(255)), schema="ras_ci")
+    op.add_column("instrument", sa.Column("file_length", Integer), schema="ras_ci")
+
+
+def downgrade():
+    pass
```
insacuri, line 25: The downgrade should be pretty simple to do surely?  Might as well if it's easy enough (though it's not likely we'll ever need it)

migrations/versions/d975ea83b712_add_seft_ci_file_location.py
```diff
@@ -0,0 +1,25 @@
...
+Revises: ccd22863e633
+Create Date: 2021-09-06 10:40:27.089307
+
+"""
+import sqlalchemy as sa
+from alembic import op
+from sqlalchemy.types import Integer, String
+
+# revision identifiers, used by Alembic.
+revision = "d975ea83b712"
```
insacuri, line 13: where did this id come from?  were they copied from somewhere else or did you have to create it?

insacuri (changes requested):
Not fully finished with the review yet but theres a few starting comments that need sorting.

application/controllers/collection_instrument.py
```diff
@@ -85,20 +86,29 @@ def upload_instrument(self, exercise_id, file, ru_ref=None, classifiers=None, se
         :return: a collection instrument instance
         """
 
+        try:
+            seft_ci_bucket = GoogleCloudSEFTCIBucket(current_app.config)
+            path = seft_ci_bucket.upload_file_to_bucket(file=file)
+        except Exception as e:
+            log.error("An error occurred when trying to put SEFT CI in bucket")
+            log.error(e)
```
warrenbailey, line 94: You could .exception here as that automatically logs the error stack trace. e.g.
```
```

application/controllers/collection_instrument.py
```diff
@@ -410,6 +402,23 @@ def get_instruments_by_exercise_id_csv(exercise_id, session=None):
...
+                csv += csv_format.format(
+                    count=count,
+                    file_name=instrument.file_location,
+                    length=len(file),
+                    date_stamp=instrument.stamp,
+                )
+                count += 1
+            return csv
+        except Exception as e:
+            log.error("Couldn't find SEFT CI from bucket")
```
warrenbailey, line 419: as above, .exception

application/controllers/collection_instrument.py
```diff
@@ -453,6 +462,15 @@ def get_instrument_data(instrument_id, session):
 
         data = None
         file_name = None
+
+        try:
+            seft_ci_bucket = GoogleCloudSEFTCIBucket(current_app.config)
+            file = seft_ci_bucket.download_file_from_bucket(instrument.file_location)
+            return file, instrument.file_location
+        except Exception as e:
+            log.error("Couldn't find SEFT CI in GCP bucket; will try database instead")
```
warrenbailey, line 471: and here

application/models/google_cloud_bucket.py
```diff
@@ -0,0 +1,34 @@
...
+log = structlog.wrap_logger(logging.getLogger(__name__))
+
+
+class GoogleCloudSEFTCIBucket:
+    def __init__(self, config):
+        self.project_id = config["GOOGLE_CLOUD_PROJECT"]
+        self.bucket_name = config["SEFT_CI_BUCKET_NAME"]
+        self.client = storage.Client(project=self.project_id)
+        self.bucket = self.client.bucket(self.bucket_name)
+        self.prefix = config["SEFT_CI_BUCKET_FILE_PREFIX"]
```
warrenbailey, line 15: There's one thing missing from this, we need to manage the keys so we can do customer supplied keys with google. If you refer to the design it is detailed on there.

application/models/google_cloud_bucket.py
```diff
@@ -0,0 +1,34 @@
...
+        else:
+            path = file.filename
+        blob = self.bucket.blob(path)
+        blob.upload_from_file(file.stream)
+        log.info("Successfully put SEFT CI in bucket")
+
+        return path
+
+    def download_file_from_bucket(self, file_location):
+        log.info("Downloading SEFT CI from GCP bucket")
```
warrenbailey, line 30: might be useful to log out the file location

application/controllers/collection_instrument.py
```diff
@@ -453,6 +462,15 @@ def get_instrument_data(instrument_id, session):
 
         data = None
         file_name = None
+
```
warrenbailey, line 465: I think it might be better to query the database result for a file location and if that exists, then go get the file from the bucket and if it doesn't exist then load the data from the database as it currently does. 

At the minute you're trying to retrieve everything from the bucket, then catching and logging the error and if that fails then getting the db result. 

At first most CIs will be retrieved from the database so this will result in a lot of errors in our logs, when they're not really errors.

application/controllers/collection_instrument.py
```diff
@@ -410,6 +402,23 @@ def get_instruments_by_exercise_id_csv(exercise_id, session=None):
...
+                file = seft_ci_bucket.download_file_from_bucket(instrument.file_location)
+                csv += csv_format.format(
+                    count=count,
+                    file_name=instrument.file_location,
+                    length=len(file),
+                    date_stamp=instrument.stamp,
+                )
+                count += 1
+            return csv
+        except Exception as e:
```
warrenbailey, line 418: same here too, lets do the database check first

warrenbailey (changes requested):

warrenbailey:
As discussed lets feature flag this

application/models/google_cloud_bucket.py
warrenbailey: As discussed we'll spin this out into a separate card as it's going in feature flagged off

application/controllers/gcp_survey_response.py
```diff
@@ -46,7 +46,7 @@ def __init__(self, config):
 
         # Pubsub config
         self.publisher = None
-        self.seft_pubsub_project = self.config["SEFT_PUBSUB_PROJECT"]
+        self.seft_pubsub_project = self.config["GOOGLE_CLOUD_PROJECT"]
```
warrenbailey, line 49: I think this needs to stay, as the pubsub topic is in a different project.

_infra/helm/collection-instrument/templates/deployment.yaml
```diff
@@ -191,9 +191,15 @@ spec:
             value: "{{ .Values.gcp.bucket.name }}"
           - name: SEFT_BUCKET_FILE_PREFIX
             value: "{{ .Values.gcp.bucket.prefix }}"
-          - name: SEFT_PUBSUB_PROJECT
```
warrenbailey, line 194: i don't think we should be removing this?

_infra/helm/collection-instrument/templates/deployment.yaml
warrenbailey: I think we need it though, as the pub sub topic for SDX is in a different project

warrenbailey (changes requested):
still need to review, but this stood out to me straight away

application/controllers/collection_instrument.py
```diff
@@ -453,7 +460,17 @@ def get_instrument_data(instrument_id, session):
 
         data = None
         file_name = None
+
         if instrument:
+            if instrument.file_location is not None:
```
warrenbailey, line 465: I think you could just do `if instrument.file_location` as None is falsy in pyton

application/controllers/collection_instrument.py
```diff
@@ -453,7 +460,17 @@ def get_instrument_data(instrument_id, session):
...
+                if current_app.config["SEFT_CI_DATABASE_TABLE_DEPRECATED"] is True:
```
warrenbailey, line 466: likewise this could be simplify to 
```
 if current_app.config["SEFT_CI_DATABASE_TABLE_DEPRECATED"]
```

warrenbailey (changes requested):

application/controllers/collection_instrument.py
```diff
@@ -85,20 +86,30 @@ def upload_instrument(self, exercise_id, file, ru_ref=None, classifiers=None, se
...
+            try:
+                seft_ci_bucket = GoogleCloudSEFTCIBucket(current_app.config)
+                path = seft_ci_bucket.upload_file_to_bucket(file=file)
+            except Exception:
+                log.exception("An error occurred when trying to put SEFT CI in bucket")
+
         log.info("Upload exercise", exercise_id=exercise_id)
 
         validate_uuid(exercise_id)
         instrument = InstrumentModel(ci_type="SEFT")
```
warrenbailey, line 100: I think the next two lines need to stay if we're allowing the feature to be toggled on or off, otherwise the old database stuff won't work

application/controllers/collection_instrument.py
```diff
@@ -354,24 +365,6 @@ def _find_or_create_business(ru_ref, session):
             business = BusinessModel(ru_ref=ru_ref)
         return business
 
-    @staticmethod
-    def _create_seft_file(instrument_id, file):
-        """
```
warrenbailey, line 359: this needs to stay temporarily otherwise the old database mechanism will no longer work

application/controllers/collection_instrument.py
```diff
@@ -414,12 +407,26 @@ def get_instruments_by_exercise_id_csv(exercise_id, session=None):
             return None
 
         for instrument in exercise.instruments:
-            csv += csv_format.format(
-                count=count,
-                file_name=instrument.name,
-                length=instrument.seft_file.len if instrument.seft_file else None,
-                date_stamp=instrument.stamp,
-            )
+            if instrument.file_location is not None:
```
warrenbailey, line 410:
```
    if instrument.file_location
```

application/controllers/collection_instrument.py
```diff
@@ -414,12 +407,26 @@ def get_instruments_by_exercise_id_csv(exercise_id, session=None):
...
+                if current_app.config["SEFT_CI_DATABASE_TABLE_DEPRECATED"] is True:
```
warrenbailey, line 411:
```
if current_app.config["SEFT_CI_DATABASE_TABLE_DEPRECATED"]
```

warrenbailey (changes requested):

application/controllers/collection_instrument.py
warrenbailey: and wrapped in the feature flag accordingly

warrenbailey:
The feature flag doesn't work for me? Regardless if I set it to true / false it always goes to the bucket

application/controllers/collection_instrument.py
```diff
@@ -85,20 +87,34 @@ def upload_instrument(self, exercise_id, file, ru_ref=None, classifiers=None, se
...
+                log.exception("An error occurred when trying to put SEFT CI in bucket")
+
         log.info("Upload exercise", exercise_id=exercise_id)
 
         validate_uuid(exercise_id)
         instrument = InstrumentModel(ci_type="SEFT")
 
-        seft_file = self._create_seft_file(instrument.instrument_id, file)
-        instrument.seft_file = seft_file
+        if current_app.config["SEFT_CI_DATABASE_TABLE_DEPRECATED"] is False:
```
warrenbailey, line 102: This won't work because it'll evaluate to 
```
if "false" is False
```
Which is False and not true

warrenbailey: You'll need to use the strtobool function to convert it to a boolean

https://github.com/ONSdigital/response-operations-ui/blob/57f78176c565f0d72acbfe2fcf69007831139f0b/config.py#L23

warrenbailey (changes requested):

warrenbailey:
I think it would be more elegant to have the feature flag check in a single place for both upload and download, rather than being scattered through the code in multiple places, as it makes it tricky to follow what is happening.

What I would have done, is to have a function for uploading to GCS and a function to save to the database. Then two more for downloading from GCS and downloading from the database. 

Then in the upload and download views functions, i'd add the feature flag check there and send the code down the associate path. It might result in a bit of duplicated code temporarily but at a later date we can just remove the upload/download from database functions completely.

application/controllers/collection_instrument.py
warrenbailey: Also this is quite hard to read as it's a double negative.

The feature flag might be better as SEFT_GCS_ENABLED. Then you could do
```
 if not current_app.config["SEFT_GCS_ENABLED"]:
```

which to me seems much easier to read

warrenbailey (changes requested):
This is much better now, it's easier to follow the code with the flag top level and two distinct code paths. Great work :)

However there's still a few things that need to be sorted before this can be merged. Firstly that config needs to be fix, it's a little confusing as we publish to a topic in an SDX project but we can't have a gcpDev section like that. We'll need to think how we separate this in a clear way.

I also missed this the last few times I've reviewed, but you've put the file location column on the wrong table. It should be on the SEFT instrument table. Basically if the file exists in the database then the data field should be populated and the file location column should be null but if the file exists in the bucket then file location column should be populated and the data column blank. I think you can then do away with the new file_length column completely.

If you read my design it details the database columns on there and I as actually thinking all you need is a flag indicating if its in gcp, as you already have the file name on that table and therefore all the information you need to construct the file location.

Finally we need to maintain backwards compatibility, so with the flag enabled you should be able to download files from both the database and the bucket depending on where they exists. At the minute, if you upload the file to the database and then enable the flag the system then errors when the user attempts to download the CI.

insacuri:
/deploy manna configBranch=adding-collection-instrument-seft-ci-bucket-stuff

insacuri:
/deploy manna configBranch=adding-collection-instrument-seft-ci-bucket-stuff

anwilkie:
/deploy wilkia configBranch=adding-collection-instrument-seft-ci-bucket-stuff

Makefile
```diff
@@ -8,7 +8,7 @@ build-kubernetes:
 	docker build -f _infra/docker/Dockerfile .
 
 lint:
-	pipenv check ./application ./tests
+	#pipenv check ./application ./tests
```
warrenbailey, line 11: is this intentional?

application/controllers/collection_instrument.py
```diff
@@ -414,12 +461,25 @@ def get_instruments_by_exercise_id_csv(exercise_id, session=None):
...
+                    seft_ci_bucket = GoogleCloudSEFTCIBucket(current_app.config)
+                    file = seft_ci_bucket.download_file_from_bucket(instrument.seft_file.file_name)
+                    csv += csv_format.format(
+                        count=count,
+                        file_name=instrument.seft_file.file_name,
+                        length=len(file),
+                        date_stamp=instrument.stamp,
+                    )
+                except Exception:
+                    log.exception("Couldn't find SEFT CI from bucket")
```
warrenbailey, line 475: really minor, but should read "Couldn't find SEFT CI in bucket"

config.py
```diff
@@ -19,11 +20,19 @@ class Config(object):
...
-    SEFT_PUBSUB_PROJECT = os.getenv("SEFT_PUBSUB_PROJECT", "ras-rm-sandbox")
-    SEFT_PUBSUB_TOPIC = os.getenv("SEFT_PUBSUB_TOPIC", "ras-rm-notify-test")
+    SEFT_UPLOAD_BUCKET_FILE_PREFIX = os.getenv("SEFT_UPLOAD_BUCKET_FILE_PREFIX")
+
+    # In preprod and prod, this project will point to the SDX project as they own the topic.  In dev environments
+    # this will be our ras-rm project as we create the topic ourselves.
+    SEFT_UPLOAD_PROJECT = os.getenv("SEFT_UPLOAD_PROJECT", "ras-rm-sandbox")
+    SEFT_UPLOAD_PUBSUB_TOPIC = os.getenv("SEFT_UPLOAD_PUBSUB_TOPIC", "ras-rm-notify-test")
+
+    COLLECTION_INSTRUMENT_BUCKET_NAME = os.getenv("COLLECTION_INSTRUMENT_BUCKET_NAME")
```
warrenbailey, line 32: This is fine as it is, but I wonder if 'SEFT_DOWNLOAD_BUCKET_NAME' might be clearer, as it's a bucket for seft ci's only.

warrenbailey (changes requested):
Looks really good now, a few very minor things and we're good to go

insacuri (approved):

warrenbailey (approved):

PR #255, opened 2021-10-07
Seft ci subfolder stuff

anwilkie:
# What and why?
The SEFT CIs need to be stored in subfolders in the format (survey-ref)/(collection-exercise-id)

# How to test?
Switch feature flag on
Upload SEFT CI and check that it's been put into the correct subfolder
Then, make sure you can download same file from frontstage

# Trello
[Card](https://trello.com/c/Zcz9wPxW)

### DIFF
--- a/application/controllers/collection_instrument.py
+++ b/application/controllers/collection_instrument.py
@@ -154,8 +154,9 @@ class CollectionInstrument(object):
             seft_ci_bucket = GoogleCloudSEFTCIBucket(current_app.config)
             seft_ci_bucket.upload_file_to_bucket(file=file)
             instrument.seft_file.gcs = True
-        except Exception:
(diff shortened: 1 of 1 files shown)


application/controllers/collection_instrument.py
```diff
@@ -155,7 +155,7 @@ def upload_to_bucket(self, exercise_id, file, ru_ref=None, classifiers=None, ses
             seft_ci_bucket.upload_file_to_bucket(file=file)
             instrument.seft_file.gcs = True
         except Exception:
-            log.exception("An error occurred when trying to put SEFT CI in bucket")
+            log.error("An error occurred when trying to put SEFT CI in bucket")
```
warrenbailey, line 158: I think you miss understood my comment, this should stay as exception but also raise the error. I.e. 

```
  except Exception as e:
            raise e
```

PR #256, opened 2021-10-07
RAS SEFT upload files customer supplied key

creativeappdevamit:
# What and why?

# How to test?

# Trello
https://trello.com/c/4S050Jyn/942-store-seft-ci-encrypted-in-folders

### DIFF
--- a/application/controllers/collection_instrument.py
+++ b/application/controllers/collection_instrument.py
@@ -132,13 +132,9 @@ class CollectionInstrument(object):
 
         seft_file = self._create_seft_file(instrument.instrument_id, file, encrypt_and_save_to_db=False)
         instrument.seft_file = seft_file
-
-        try:
-            seft_ci_bucket = GoogleCloudSEFTCIBucket(current_app.config)
-            seft_ci_bucket.upload_file_to_bucket(file=file)
-            instrument.seft_file.gcs = True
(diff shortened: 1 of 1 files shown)


application/controllers/collection_instrument.py
```diff
@@ -132,13 +132,9 @@ def upload_to_bucket(self, exercise_id, file, ru_ref=None, classifiers=None, ses
...
         seft_file = self._create_seft_file(instrument.instrument_id, file, encrypt_and_save_to_db=False)
         instrument.seft_file = seft_file
-
-        try:
-            seft_ci_bucket = GoogleCloudSEFTCIBucket(current_app.config)
-            seft_ci_bucket.upload_file_to_bucket(file=file)
-            instrument.seft_file.gcs = True
-        except Exception:
-            log.exception("An error occurred when trying to put SEFT CI in bucket")
+        seft_ci_bucket = GoogleCloudSEFTCIBucket(current_app.config)
```
warrenbailey, line 135: I see why you've removed the error handling, but can the client raise other exceptions? Should we be converting them to RasErrors ?

creativeappdevamit: the api will raise an error as we were handling it, it never stoped the process
we want to stop and respond on ROps

PR #265, opened 2021-12-02
read key file in as binary instead of ASCII

anwilkie:
# What and why?
The crypto key needs to be read in as binary instead of being converted to ASCII like it once did.

# How to test?
Change the crypto keys in the secret manager on GCP to binary(???)

# Trello
[Card](https://trello.com/c/tHpI5VMv)


insacuri:
This functionality is in frontstage now so theres no point

PR #270, opened 2022-02-25
Start of migration work

insacuri:
# What and why?

Currently new instruments are saved to the bucket, but older ones are still in the database.  This PR adds an endpoint to get the data, decrypt it, then upload it to the bucket.  It then sets the gcs flag to true against the instrument so this service can know where to find it.

# How to test?

- Deploy this PR to your environment
- Edit the deployment in k8s and update SEFT_GCS_ENABLED to false.  This will save any set up collection instruments to the database instead of the bucket.
- Run the SEFT acceptance tests, verify the data is in the database and the gcs flag is false.
- Find the instrument id for the instrument you want to migrate
- Add the id to the migration script in this PR.  Then run it by doing `pipenv run python migrate_instruments_to_gcp.py`
-  It should work and you should be able to see it in gcs in your namespace.  Also the gcs flag should be true against the instrument in question.
- Try and download the SEFT file and open it as a respondent.

Note the data is still in the database after this process.  Once we've migrated them all to the bucket, and left it a few days to verify that everything is okay, we can run some SQL to delete all the `data` values from the database.

# Trello

### DIFF
--- a/application/controllers/sql_queries.py
+++ b/application/controllers/sql_queries.py
@@ -2,6 +2,7 @@ from application.models.models import (
     BusinessModel,
     ExerciseModel,
     InstrumentModel,
+    SEFTModel,
     SurveyModel,
 )
 
@@ -24,3 +25,7 @@ def query_instrument_by_id(instrument_id, session):
 
 def query_instrument(session):
     return session.query(InstrumentModel)
+
+
+def query_seft_instrument_by_instrument_id(instrument_id, session):
+    return session.query(SEFTModel).filter(SEFTModel.instrument_id == instrument_id).one()
--- a/scripts/migrate_instruments_to_gcp.py
+++ b/scripts/migrate_instruments_to_gcp.py
@@ -1,0 +1,34 @@
+import sys
+
+import requests
+from requests.exceptions import HTTPError
+
+# Run this script by doing `pipenv run python scripts/migrate_instruments_to_gcp.py`
+
+# Populate this with the instrument ids you wish to migrate
+instrument_ids = []
+
+print("Beginning migrating instruments")
+
+for instrument_id in instrument_ids:
+    print(f"Migrating instrument [{instrument_id}")
+    # If you're port forwarding, this should remain the same regardless of environment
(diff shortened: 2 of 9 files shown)


application/controllers/sql_queries.py
```diff
@@ -24,3 +25,7 @@ def query_instrument_by_id(instrument_id, session):
 
 def query_instrument(session):
     return session.query(InstrumentModel)
+
+
+def query_seft_instrument_by_instrument_id(instrument_id, session):
```
creativeappdevamit, line 30: is one() required? wont it only send one record back?

scripts/migrate_instruments_to_gcp.py
```diff
@@ -0,0 +1,33 @@
...
+
+# Populate this with the instrument ids you wish to migrate
+instrument_ids = []
+
+print("Beginning migrating instruments")
+
+for instrument_id in instrument_ids:
+    print(f"Migrating instrument [{instrument_id}")
+    # If you're port forwarding, this should remain the same regardless of environment
+    url = f"http://localhost:8002/collection-instrument-api/1.0.2/migrate/{instrument_id}"
```
creativeappdevamit, line 16: localhost?

creativeappdevamit (changes requested):

creativeappdevamit:
Are we planning to have some unit test for this?

application/controllers/sql_queries.py
insacuri: I want it to throw an exception if it somehow got more than one back.  I don't think it's possible but theres no harm in the extra check

scripts/migrate_instruments_to_gcp.py
insacuri: Yeah, when you run the script you have to set up port forwarding to the collection instrument pod.   It's a one time use script so I didn't put too much into making it configurable.  Just change the port to whatever you need it to be and go from there

insacuri:
I wasn't planning on it as once the migration has been done once then we'll be deleting all of this as it's a one time operation

creativeappdevamit (approved):

PR #276, opened 2022-05-13
Remove SEFT_GCS_ENABLED flag

insacuri:
# What and why?

All of our instruments are now saved and loaded from a gcs bucket rather then encrypted in the database.  This functionality has been working for a while now so we can safely remove the feature toggle as we have no intention of ever turning it on again.

# How to test?

# Trello


application/views/collection_instrument_view.py
```diff
@@ -31,10 +31,7 @@ def before_collection_instrument_view():
 def upload_collection_instrument(exercise_id, ru_ref=None):
     file = request.files["file"]
     classifiers = request.args.get("classifiers")
-    if current_app.config["SEFT_GCS_ENABLED"] is False:
-        instrument = CollectionInstrument().upload_instrument(exercise_id, file, ru_ref=ru_ref, classifiers=classifiers)
-    else:
-        instrument = CollectionInstrument().upload_to_bucket(exercise_id, file, ru_ref=ru_ref, classifiers=classifiers)
+    instrument = CollectionInstrument().upload_to_bucket(exercise_id, file, ru_ref=ru_ref, classifiers=classifiers)
```
LJBabbage, line 34: Not for this PR, as you haven't changed it, but because the highlight made me see it, but the instantiation/creation and the upload should be separate, the method upload_to_bucket should not really be returning an instrument. The naming would suggest an action returning bool

tests/views/test_collection_instrument_view.py
```diff
@@ -28,6 +28,7 @@
 linked_exercise_id = "fb2a9d3a-6e9c-46f6-af5e-5f67fec3c040"
 url_collection_instrument_link_url = "http://localhost:8145/collection-instrument/link"
 url_survey_url = "http://localhost:8080/surveys/cb0711c3-0ac8-41d3-ae0e-567e5ea1ef87"
+survey_response_json = {"surveyId": "cb0711c3-0ac8-41d3-ae0e-567e5ea1ef87", "surveyRef": "139"}
```
LJBabbage, line 31: We have made a var in some instances, and others we have written inlined, any reason for that? Also find it slightly odd they have different UUID's but both are returning  "surveyRef": "139", do we need both?

tests/views/test_collection_instrument_view.py
```diff
@@ -154,6 +156,11 @@ def test_upload_collection_instrument_if_survey_does_not_exist(self):
     @requests_mock.mock()
     def test_collection_instrument_upload_with_ru(self, mock_bucket, mock_request):
         mock_request.post(url_collection_instrument_link_url, status_code=200)
+        mock_request.get(
+            "http://localhost:8080/surveys/db0711c3-0ac8-41d3-ae0e-567e5ea1ef87",
+            status_code=200,
+            json={"surveyId": "db0711c3-0ac8-41d3-ae0e-567e5ea1ef87", "surveyRef": "139"},
+        )
```
LJBabbage, line 163: This chunk is repeated 3 times, we should try to extract this. This is also what I meant about writing inline as well

application/views/collection_instrument_view.py
insacuri: I can make a card to improve it, as I think you have a point, but this PR is just reorganising code and removing a flag.  I don't want to increase the scope of it unless there was a really good reason for it.

tests/views/test_collection_instrument_view.py
insacuri: The reason?  Not a clue.  Why do we need both?  Equally no idea.  I made some small changes because the test broke but it probably wouldn't hurt to polish this up whist I'm here as it's a small amount of effort

LJBabbage (approved):

PR #282, opened 2022-08-08
Delete SEFT collection instrument

LJBabbage:
# What and why?
This PR adds an end point to delete SEFT collection instruments. I have confirmed with Gareth that SEFTs should not be unlinked now and should always been deleted, so have updated the code as appropriate. 

N.B I have done a few minor refactoring changes some of the test were adding in extra data and not reusing vars
# How to test?
The best way is by using this [PR](https://github.com/ONSdigital/response-operations-ui/pull/758#pullrequestreview-1066603876) whilst this one is deployed to GCP. Add and remove collection instruments making sure they are added and removed from both the db and GCP (your instance in ras-rm-seft-ci-dev found in Cloud Storage) 

N.B there are 2 general types of SEFT one business specific which will create a record in the business table and one won't
they both need checking. I used the file names 064_201803_0001.xlsx (not specific to a business) and 11111111111.xlsx (for business specific, basically an 11 digit number)

# Trello


application/controllers/collection_instrument.py
```diff
@@ -277,17 +278,43 @@ def unlink_instrument_from_exercise(self, instrument_id, exercise_id, session=No
             bound_logger.info("Failed to unlink, unable to find instrument or exercise")
             raise RasError("Unable to find instrument or exercise", 404)
 
-        instrument.exercises.remove(exercise)
-        for business in instrument.businesses:
```
LJBabbage, line 281: Removed as per the comment, unlinking is not something a SEFT should be doing (confirmed with Gareth) and as instrument businesses is a SEFT association it becomes redundant

application/controllers/collection_instrument.py
```diff
@@ -277,17 +278,43 @@ def unlink_instrument_from_exercise(self, instrument_id, exercise_id, session=No
...
-            bound_logger.info("Removing business/exercise link", business_id=business.id, ru_ref=business.ru_ref)
-            business.instruments.remove(instrument)
+        if instrument.type == "SEFT":
+            # SEFT instruments need to be deleted from both GCP and the db, removing just the link will leave orphaned
+            # data. When deleting SEFT instruments, the link will automatically be removed without this function
+            raise RasError(
+                f"{instrument_id} is of type SEFT which should be deleted and not unlinked",
+                405,
```
LJBabbage, line 286: Debatable on the 405

application/controllers/collection_instrument.py
```diff
@@ -277,17 +278,43 @@ def unlink_instrument_from_exercise(self, instrument_id, exercise_id, session=No
...
+        if not instrument:
+            raise RasError(f"Collection instrument {instrument_id} not found", 404)
+        if instrument.type != "SEFT":
+            raise RasError(
+                f"Only SEFT collection instruments can be deleted {instrument_id} has type {instrument.type}",
+                405,
+            )
+
+        session.delete(instrument)
+        gcs_seft_bucket = GoogleCloudSEFTCIBucket(current_app.config)
```
LJBabbage, line 314: It's debatable whether this should raise an exception if it fails, I have left as is, but if it can't find the blob is just writing an error in the logs enough?

application/controllers/collection_instrument.py
```diff
@@ -277,17 +278,43 @@ def unlink_instrument_from_exercise(self, instrument_id, exercise_id, session=No
...
+        file_path = self._build_seft_file_path(instrument)
+        gcs_seft_bucket.delete_file_from_bucket(file_path)
```
LJBabbage, line 316: It's debatable whether this should raise an exception if it fails, I have left as is, but if it can't find the blob is just writing an error in the logs enough?

application/controllers/collection_instrument.py
JacobHarrisonONS: 405 makes sense, do we want the error (depending how often this is attempted against something not valid) if thats not an issue then fine

application/controllers/collection_instrument.py
JacobHarrisonONS: Is the file not existing an issue ? it should be very rare the file not being there to delete. I think the unlinking / deleting should be enough since there would be the asumption that if they were correcting a mistake a few file would be loaded.

application/controllers/collection_instrument.py
LJBabbage: It should feasibly never happen as it would need someone to either hit the end point or mess around with the UI, which is highly unlikely considering it's internal. I think it's still good pratice to do that, but I can see it's not a pattern we do much in RAS/RM. One for tech maybe, as well as how we handle api calls between services

application/controllers/collection_instrument.py
LJBabbage: I ummed and arg on that one as well, the delete_file_from_bucket writes out to the logs which is at least something I guess. I came to the conclusion that I didn't know the current state of the data so it is completely feasible to have a record in the table and not in GCS. I also thought as long as the instrument is removed from the db and thus the UI we can handle orphaned data in GCS if we have to.

JacobHarrisonONS (approved):

PR #283, opened 2022-08-10
Remove more unused columns

mfcrocker:
# What and why?
All SEFT CIs are now in GCS - there's no need to have a flag on this any more

# How to test?

# Trello
https://trello.com/c/ABWCOijG/1882-delete-instrument-and-gcs-columns-from-seftinstrument-table-in-ras-collection-instrument


_infra/helm/collection-instrument/Chart.yaml
```diff
@@ -14,8 +14,8 @@ type: application
 
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
-version: 2.1.36
+version: 2.1.37
```
insacuri, line 17: I think this could arguably be 2.2.0 or even 3.0.0 as theres no backwards compatibility after we do this change

application/controllers/collection_instrument.py
```diff
@@ -500,21 +488,15 @@ def get_instrument_data(instrument_id, session):
...
-                    exercise_id = str(instrument.exids[0])
-                    file_path = survey_ref + "/" + exercise_id + "/" + instrument.seft_file.file_name
-                    seft_ci_bucket = GoogleCloudSEFTCIBucket(current_app.config)
-                    file = seft_ci_bucket.download_file_from_bucket(file_path)
-                    return file, instrument.seft_file.file_name
-                except Exception:
-                    log.exception("Couldn't find SEFT CI in GCP bucket")
-            else:
-                log.info("Decrypting collection instrument data", instrument_id=instrument_id)
-                cryptographer = Cryptographer()
```
insacuri, line 515: Just looking at the repo after this change, the cryptographer isn't used anymore.  Can you remove it and the stuff that goes with it as well?

insacuri (changes requested):

application/controllers/collection_instrument.py
insacuri: You can remove the ons_cryptokey value too from the config and helm chart as well.  I think the secret might have to stay where it is as frontstage is using it I think, but I could be wrong about that.

PR #284, opened 2022-09-15
Remove ru_ref and collection_exercise data from classifiers in /collectioninstrument endpoint

mfcrocker:
# What and why?

The BRES collection exercise was slow to load as it's the only one that's had to load 2.5k instruments.  This highlighted some inefficiencies in the code.  One thing that was noticed was that there is an inefficient call to get the ru_ref and collection exercise ids and add it to the payload.   After looking at all the calls in the system, it was concluded that nothing actually uses this data so we'll leave it blank as it makes a 2.5k instrument exercise go from 25 seconds to load down to 10 seconds.

If it passes our testing and goes live, then we can do another pass to remove the 'classifier' key entirely from the returned dictionary.

# How to test?

- Deploy PR to environment
- Run acceptance tests to seed data
- Create, set ready for live and sign up for the following collection exercises with multiple formtypes:
  - any EQ
  - any SEFT, except BRES
  - a BRES survey (for this, 'multiple formtypes' means 2+ reporting units that will get unique collection instruments)

# Trello

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,4 +8,4 @@ WORKDIR /app
 COPY . /app
 RUN pipenv install --deploy --system
 EXPOSE 8002
-CMD ["gunicorn", "-b", "0.0.0.0:8080", "--workers", "8", "--worker-class", "gevent", "--worker-connections", "1000", "--timeout", "30", "--keep-alive", "2", "app:app"]
+CMD ["gunicorn", "-b", "0.0.0.0:8080", "--workers", "6", "--worker-class", "gevent", "--worker-connections", "1000", "--timeout", "30", "--keep-alive", "2", "app:app"]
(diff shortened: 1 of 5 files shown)


LJBabbage (commented):
Not sure about this change, I can see a few places which uses COLLECTION_EXERCISE https://github.com/ONSdigital/response-operations-ui/blob/main/response_operations_ui/controllers/collection_instrument_controllers.py#L203-L213, which would bubble up here https://github.com/ONSdigital/response-operations-ui/blob/main/response_operations_ui/views/collection_exercise.py#L76 meaning anything that uses ce_details["collection_instruments"] might not work as expected which I can see a few places in this repo. I think if we are doing this we need to eliminate the query builds that use searchString with COLLECTION_EXERCISE and RU_REF first

LJBabbage (approved):
Taking a pragmatic view that the search params are updated/removed in another card then this can be merged. I have done a light test in dev and it did work

PR #289, opened 2022-10-28
Investigate bres slowdown

insacuri:
# What and why?

When uploading collection instruments for an exercise, we were finding that once the number of instruments reached >500, it would take noticeably longer to add any subsequent ones.  As an example, for the biggest exercise, it took about 10 seconds to upload a new one when it got to ~2600 instruments.

This was happening in the duplicate filename check, where it checks to make sure a duplicate instrument isn't being uploaded.  It was taking about 4ms per record as it needed to go back to the database each time it wanted the filename of the seft record.  4ms isn't long, but when multiplied by 1000+ it gets slow quickly!.

This PR makes the loading more eager, so when you call `exercises.instruments`, it not only loads the record for the instruments, but it also loads the SEFT records at the same time.  Then when we loop through them, instead of going to the database for each record (slow) we're just iterating over a list (much faster!)

# How to test?

- Deploy to the envrionment
- Run acceptance tests (both EQ and SEFT)
- Test a BRES collection exercise (both setting ready for live and submitting one)

You can also deploy this change into prod and test it on a BRES survey with >2600 instruments.  Uploading should take just under a second, rather then the 10 seconds from before.

# Trello


application/controllers/collection_instrument.py
```diff
@@ -125,19 +125,22 @@ def upload_to_bucket(self, exercise_id, file, ru_ref=None, classifiers=None, ses
     @staticmethod
     def validate_non_duplicate_instrument(file, exercise_id, session):
         exercise = query_exercise_by_id(exercise_id, session)
-
+        log.info("Validating if instrument is already uploaded for this exercise", exercise_id=exercise_id)
```
LJBabbage, line 128: Just a bit too much logging, there is so much across our repo I don't think we need function entry

LJBabbage (approved):
Works fine, I just don't believe we need the extra logging, There is so much already and logging on success seems unnecessary.

PR #297, opened 2023-02-22
Add collection instrument form type check eq and seft

LJBabbage:
# What and why?
This PR validates if a collection instrument can be uploaded according to it's form_type. 
# How to test?
Test with this PR deploy as well https://github.com/ONSdigital/response-operations-ui/pull/844
Using this PR create a EQ_AND_SEFT survey in rops, upload an eQ form_type (i.e 0001) then try to upload a SEFT using the same form type (a file named 064_201803_0001 would do it), it should return a warning. Repeat the process but the other way around Upload a SEFT first then an eQ. This is survey wide, so try across CE as well (i.e create a CE and upload a eQ, then in a different CE try to add a SEFT with the same form_type)
# Trello

### DIFF
--- a/application/controllers/collection_instrument.py
+++ b/application/controllers/collection_instrument.py
@@ -8,7 +8,7 @@ from sqlalchemy.orm import Session
 from application.controllers.helper import validate_uuid
 from application.controllers.service_helper import (
     collection_instrument_link,
-    get_survey_ref,
+    get_survey_service_details,
     service_request,
 )
 from application.controllers.session_decorator import with_db_session
@@ -17,6 +17,7 @@ from application.controllers.sql_queries import (
     query_exercise_by_id,
     query_instrument,
     query_instrument_by_id,
+    query_instruments_form_type_with_different_survey_mode,
     query_survey_by_id,
 )
 from application.exceptions import RasError
@@ -76,9 +77,9 @@ class CollectionInstrument(object):
         return result
 
     @with_db_session
-    def upload_to_bucket(self, exercise_id, file, ru_ref=None, classifiers=None, session=None):
+    def upload_seft_to_bucket(self, exercise_id, file, ru_ref=None, classifiers=None, session=None):
         """
-        Encrypt and upload a collection instrument to the bucket and db
+        Encrypt and upload a SEFT collection instrument to the bucket and db
 
         :param exercise_id: An exercise id (UUID)
         :param ru_ref: The name of the file we're receiving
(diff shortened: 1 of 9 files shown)


application/controllers/collection_instrument.py
```diff
@@ -87,42 +88,50 @@ def upload_to_bucket(self, exercise_id, file, ru_ref=None, classifiers=None, ses
...
-        seft_file = self._create_seft_file(instrument.instrument_id, file)
-        instrument.seft_file = seft_file
-
-        exercise = self._find_or_create_exercise(exercise_id, session)
-        instrument.exercises.append(exercise)
 
         survey = self._find_or_create_survey_from_exercise_id(exercise_id, session)
-        instrument.survey = survey
+        survey_id = survey.survey_id
+        survey_service_details = get_survey_service_details(survey_id)
```
LJBabbage, line 98: I moved the order around as once past validate_uuid and validate_non_duplicate_instrument the next place you could fall over is the call to the survey service and you can't avoid it. Therefore it makes sense to call it next to save yourself doing a load of stuff that will just blow up later if it fails

application/controllers/collection_instrument.py
```diff
@@ -222,17 +231,22 @@ def upload_instrument_with_no_collection_exercise(self, survey_id, classifiers=N
...
         survey = self._find_or_create_survey_from_survey_id(survey_id, session)
+        survey_service_details = get_survey_service_details(survey.survey_id)
+        ci_type = "EQ"
+        instrument = InstrumentModel(ci_type=ci_type)
         instrument.survey = survey
-
         if classifiers:
-            deserialized_classifiers = loads(classifiers)
+            classifiers = loads(classifiers)
+            if survey_service_details["surveyMode"] == "EQ_AND_SEFT":
```
LJBabbage, line 241: This is debatable, checking the survey_mode is not necessarily needed as you could do it on everything. I included it in the SEFT one because survey_service_details was in scope and therefore an extra call was not needed. Survey service is not in scope here and I have had to get it, which begs the question should we be doing it

application/controllers/service_helper.py
```diff
@@ -4,85 +4,17 @@
 import structlog
 from flask import current_app
 
-from application.exceptions import RasError
+from application.exceptions import RasError, ServiceUnavailableException
 
 log = structlog.wrap_logger(logging.getLogger(__name__))
 
 
-def get_case_group(case_id):
```
LJBabbage, line 12: All unused so I got rid, we do need to circle back around as there are other things that can be removed

application/controllers/service_helper.py
```diff
@@ -99,19 +31,26 @@ def service_request(service, endpoint, search_value):
...
-    response.raise_for_status()
+    try:
+        response = requests.get(service_url, auth=auth)
+        response.raise_for_status()
+    except requests.HTTPError:
+        raise RasError(f"{service} returned a HTTPError")
+    except requests.ConnectionError:
+        raise ServiceUnavailableException(f"{service} returned a connection error", 503)
+    except requests.Timeout:
+        raise ServiceUnavailableException(f"{service} has timed out", 504)
```
LJBabbage, line 53: Just good practise. I had to change how survey service returned, so update the calls (with tests)

application/controllers/collection_instrument.py
```diff
@@ -76,9 +77,9 @@ def get_instrument_by_search_string(self, search_string=None, limit=None, sessio
         return result
 
     @with_db_session
-    def upload_to_bucket(self, exercise_id, file, ru_ref=None, classifiers=None, session=None):
+    def upload_seft_to_bucket(self, exercise_id, file, ru_ref=None, classifiers=None, session=None):
```
LJBabbage, line 80: Me being explicit again, this is clearly SEFT by the instrument it is setting, it's name should reflect that

application/controllers/collection_instrument.py
```diff
@@ -209,9 +218,9 @@ def validate_one_instrument_for_ru_specific_upload(exercise, business, session):
         bound_logger.info("Successfully validated there isn't an instrument for this ru for this exercise")
 
     @with_db_session
-    def upload_instrument_with_no_collection_exercise(self, survey_id, classifiers=None, session=None):
+    def upload_eq(self, survey_id, classifiers=None, session=None):
```
LJBabbage, line 221: As above, this is exclusively used for eQ, the whole with_no_collection_exercise is a red herring

tests/views/test_collection_instrument_view.py
```diff
@@ -81,26 +88,10 @@ def test_collection_instrument_upload(self, mock_bucket, mock_request):
 
         self.assertEqual(len(collection_instruments()), 2)
 
-    @mock.patch("application.controllers.collection_instrument.GoogleCloudSEFTCIBucket")
     @requests_mock.mock()
-    def test_seft_collection_instrument_delete(self, mock_bucket, mock_request):
```
LJBabbage, line 86: Just moved, I preferred to do all the uploads before a delete

application/controllers/service_helper.py
```diff
@@ -4,85 +4,17 @@
...
-
-    if response.status_code == 200:
-        collection_exercise = response.json()
-    else:
-        log.info("Collection Exercise not found", collection_exercise_id=collection_exercise_id)
-    return collection_exercise
-
-
-def get_survey_ref(survey_id):
+def get_survey_service_details(survey_id):
```
insacuri, line 12: It's a bit picky, but I think this should be either `get_survey_details` or even just `get_survey`

application/controllers/sql_queries.py
```diff
@@ -24,3 +24,16 @@ def query_instrument_by_id(instrument_id, session):
 
 def query_instrument(session):
     return session.query(InstrumentModel)
+
+
+def query_instruments_form_type_with_different_survey_mode(survey_id, form_type, survey_mode, session):
```
insacuri, line 29: Could you put a comment here so someone at a glance can know what is it you're searching for?  Something like
' Searches for instruments for a survey that match the formtype and does NOT match the supplied survey_mode'

application/controllers/collection_instrument.py
```diff
@@ -87,42 +88,57 @@ def upload_to_bucket(self, exercise_id, file, ru_ref=None, classifiers=None, ses
...
         except Exception as e:
             log.exception("An error occurred when trying to put SEFT CI in bucket")
             raise e
 
-        session.add(instrument)
         return instrument
 
+    @staticmethod
+    def validate_eq_and_seft_form_type(survey_id: str, form_type: str, survey_mode: str, session: Session) -> None:
+        instruments = query_instruments_form_type_with_different_survey_mode(survey_id, form_type, survey_mode, session)
```
insacuri, line 131: A docstring on the function would help understand what this function is doing at a glance.  Something like 'This function ensures there are no existing instruments with a given formtype for a given survey'.

I know the error text is pretty clear why the failure is happening so it might be repeating yourself but I think it'd be useful.

application/controllers/collection_instrument.py
insacuri: Good change 👍

insacuri (approved):

PR #300, opened 2023-03-02
Allow for deletion of EQ collection instruments

insacuri:
# What and why?

# How to test?

# Trello


LJBabbage (approved):
All good, even tried it with mulitmode, deliberately entered an eQ form_type that should be a SEFT. It initially refused my SEFT upload, but when I hit the delete endpoint with the eQ CI it got rid of it and then allowed my SEFT upload.

PR #301, opened 2023-03-13
Upgrade sqlalchemy 2.0

insacuri:
# What and why?

sqlalchemy 2.0 finally released so we upgraded to it to keep current.
It's not required too many changes as we used sessions and ORM models in a fairly simple way.

# How to test?

- Deploy app to environment
- Run all 4 types of acceptance tests to ensure it works
- Try a few manual things (deleting instruments, etc)

To test it with a completely fresh database
- delete the ras_ci schema
- `kubectl rollout restart deployment collection-instrument`
- run the acceptance tests again to verify that it's happy

This will likely have a few errors in the logs as all the workers will try and run alembic, and a few might fail initially because of race conditions but it all works out in the end.

# Trello


run.py
```diff
@@ -81,7 +81,8 @@ def create_database(db_connection, db_schema):
 
         if not session().query(schemata_exists).scalar():
             logger.info("Creating schema ", db_schema=db_schema)
-            engine.execute(f"CREATE SCHEMA {db_schema}")
+            session().execute(text(f"CREATE SCHEMA {db_schema}"))
```
insacuri, line 84: 2.0 doesn't let `engine.execute` anymore.  You have to do it either through a session, like the one we've created already, or through a connection.

LJBabbage (approved):
Good stuff, manually tested it, then ran all the acceptance tests, then deleted the schema and retested. I saw no issues to worry about

PR #302, opened 2023-04-14
RAS-284 Delete collection instruments by ce id

LJBabbage:
# What and why?
This PR adds an end point to delete the collection exercise and associated collection instruments on GCP

N.B 

1. As discussed in tech because of the many to many relationship between exercise and collection instruments and the state of data in PROD I can't add a cascade to exercise to remove all CIs' associated with it. This will not stop CIs' being deleted from GCP, but it will leave CI orphans in the db. This is being tackled when the CI registry is introduced when CE will be a one to many relationship

2. I have been a lot more robust with tests and error catching then how the current repo is written, however stopped short of tackling the issues with  _with_db_session_  Realistically we should have error catching and tests on all these session updates but the decorator needs a fair amount of work

3. This is only for the collection instrument services, so you will still see the CE in rops if you refresh after you call the end point

# How to test?
Deploy this PR to GCP (it is a right pain to test locally as you need to hook up the GCS bucket so far easier to deploy to GCP) and create a SEFT CE in rops and add some instruments, check they are in the bucket (ras-rm-seft-ci-dev) Look in your db and grab the CE id and then use it to hit the endpoint
```
http://localhost:8002/collection-instrument-api/1.0.2/delete/collection-exercise/<your ce id>
``` 
Use something like postman as you will need the method to be DELETE, also you will need to add in Basic auth on the request for dev its admin/secret

It's going to look a bit like this 
image

Now check that the folder and CIs are removed on GCP and that the exercise is removed from the CI DB

# Trello

### DIFF
--- a/_infra/helm/collection-instrument/Chart.yaml
+++ b/_infra/helm/collection-instrument/Chart.yaml
@@ -14,9 +14,9 @@ type: application
 
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
-version: 3.0.14
+version: 3.0.15
 
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application.
-appVersion: 3.0.14
+appVersion: 3.0.15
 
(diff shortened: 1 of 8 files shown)


SteveScorfield (approved):
Looks good to me

PR #307, opened 2023-06-05
RAS-728 Fix issue with ready for live button

LJBabbage:
# What and why?
This PR fixes the bug associated with the ready for live button not appearing after initially adding a collection instrument. I have also done some refactoring and tidying of the code connected to the problem
# How to test?
Deploy all 3 PRs, this one,  https://github.com/ONSdigital/rm-collection-exercise-service/pull/312 and https://github.com/ONSdigital/rm-collection-exercise-service/pull/312 Now on rops create an eq CE (SEFT will work for add as well, see below for bug on remove) and enter all details bar the CI. Now enter the CI and attach that CI, notice when you hit done the ready for live button appears. Go back in and remove the CI, notice it is gone

N.B there are 2 bugs
1. with EQ_AND_SEFT the current eQ adding code will remove SEFT instruments as well, a card has been create for this
2. Remove a seft instrument is not configured correctly, it currently is not calling CE, so if you add a CI and then remove that ready for live is still present. Once again this is out of scope and another card is needed
# Trello

### DIFF
--- a/application/controllers/collection_instrument.py
+++ b/application/controllers/collection_instrument.py
@@ -274,7 +274,7 @@ class CollectionInstrument(object):
     @with_db_session
     def update_exercise_eq_instruments(self, exercise_id: str, instruments: list, session=None) -> bool:
         """
-        Updates eQ instruments used by an exercise. Current instruments are used to determine which ones should be
+        Updates eQ instruments for an exercise. Current instruments are used to determine which ones should be
         appended and/or removed.
         :param exercise_id: The collection exercise id
         :param instruments: A list of instruments that the collection exercise should now have
--- a/application/controllers/collection_instrument.py
+++ b/application/controllers/collection_instrument.py
@@ -6,11 +6,7 @@ from flask import current_app
 from sqlalchemy.orm import Session
 
 from application.controllers.helper import validate_uuid
-from application.controllers.service_helper import (
-    collection_instrument_link,
-    get_survey_details,
-    service_request,
-)
+from application.controllers.service_helper import get_survey_details, service_request
 from application.controllers.session_decorator import with_db_session
 from application.controllers.sql_queries import (
     query_business_by_ru,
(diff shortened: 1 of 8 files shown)


application/controllers/collection_instrument.py
```diff
@@ -276,48 +272,34 @@ def upload_eq(self, survey_id, classifiers=None, session=None):
         return instrument
 
     @with_db_session
-    def update_collection_exercise_instruments(self, instrument_selection, exercise_id, session=None):
+    def update_exercise_eq_instruments(self, exercise_id: str, instruments: list, session=None) -> bool:
```
LJBabbage, line 275: This is only used for eQ and the name should reflect that

application/controllers/collection_instrument.py
```diff
@@ -419,36 +398,6 @@ def delete_collection_instruments_by_exercise(self, ce_id: str, session: Session
 
         return COLLECTION_EXERCISE_AND_ASSOCIATED_FILES_DELETED, 200
 
-    @staticmethod
-    def publish_remove_collection_instrument(exercise_id, instrument_id):
```
LJBabbage, line 423: Couple of issues here, 1. This shouldn't really be part of this class and has led to the issue with calls being made to external service during sessions. 2. The Collection Exercise is quite dum and cares nothing about the action or the instrument, all it is doing is running a transitionScheduleCollectionExerciseToReadyToReview, which it just needs the exercise id. With that in mind we don't need a remove or an add, or to send it the exercise or action.

application/controllers/collection_instrument.py
```diff
@@ -419,36 +398,6 @@ def delete_collection_instruments_by_exercise(self, ce_id: str, session: Session
...
-        :param exercise_id: An exercise id (UUID)
-        :param instrument_id: The id (UUID) of collection instrument
-        :return: True if message successfully sent to collection exercise
-        """
-        log.info("Publishing remove message", exercise_id=exercise_id, instrument_id=instrument_id)
-        json_message = {"action": "REMOVE", "exercise_id": str(exercise_id), "instrument_id": str(instrument_id)}
-        return collection_instrument_link(json_message)
-
-    @staticmethod
-    def publish_collection_instrument_to_collection_exercise(to_add, exercise_id, instrument_id):
```
LJBabbage, line 436: As above

application/controllers/collection_instrument.py
```diff
@@ -276,48 +272,34 @@ def upload_eq(self, survey_id, classifiers=None, session=None):
...
             instrument = self.get_instrument_by_id(instrument_id, session)
+            instrument.exercises.remove(exercise)
 
-            if to_add:
-                instrument.exercises.append(exercise_and_instruments)
-            else:
-                instrument.exercises.remove(exercise_and_instruments)
+        log.info("Collection instruments updated successfully", instruments=instruments, exercise_id=exercise_id)
 
-        self.publish_collection_instrument_to_collection_exercise(
```
LJBabbage, line 318: The cause of the bug, being in the same db_session (transaction) the publish step will effectively behaviour before any of the append remove happens, as they won't persist till the whole of the transaction has passed

tests/controllers/test_collection_instrument.py
```diff
@@ -247,22 +230,6 @@ def test_get_instrument_by_incorrect_id(self):
         # Then that instrument is not found
         self.assertEqual(instrument, None)
 
-    @requests_mock.mock()
-    def test_publish_uploaded_collection_instrument(self, mock_request):
```
LJBabbage, line 251: In the wrong place, moved

SteveScorfield (approved):
Looks good to me

PR #324, opened 2024-04-19
Upgrade requests-mock to fix idna security alert

pricem14pc:
# What and why?

To fix security alert by bumping dev package idna from 3.6 -> 3.7


arroyoAle (approved):
Changes are fine and CI still works

PR #333, opened 2024-07-23
change dockerfile directory

anwilkie:
# What and why?
The dockerfile needed to be moved to its own folder in _infra in order to make it consistent with the other repos.

A few pipenv dependencies were updated, too, and some vulnerability warnings were ignored like they are on other repos (and number 70624 was still cropping up even though flask-cors wouldn't update any further).

# How to test?
Check GA build

# Jira
[Card](https://jira.ons.gov.uk/browse/RAS-313)


Makefile
```diff
@@ -8,7 +8,7 @@ build-kubernetes:
 	docker build -f _infra/docker/Dockerfile .
 
 lint:
-	pipenv check ./application ./tests -i 51457
+	pipenv check ./application ./tests -i 70612 -i 71064 -i 70624
```
arroyoAle, line 11: There is an update for 71064 but that cane be another card

arroyoAle (approved):
Looks good and works fine. Just a small comment that would be a different card

Pipfile
```diff
@@ -25,6 +25,7 @@ pycryptodome = "*"
 requestsdefaulter = "*"
 google-cloud-storage = "*"
 google-cloud-pubsub = "*"
+urllib3 = "*"
```
matthew-robinson-ons, line 28: As with secure message we shouldn't be importing urllib3 but updating the package that manages the dependency

matthew-robinson-ons (changes requested):
As with secure message urllib3 needs looking at but the docker changes look good

_infra/docker/Dockerfile
arroyoAle, line 1: This dockerfile in the root directory is used in the `docker-compose up -d` command in the readme

arroyoAle (changes requested):
I noticed this while working on another card but if we do not have the image already built locally, the docker compose command will fail as it is expecting a docker configuration file

Pipfile
anwilkie: Fixed.

_infra/docker/Dockerfile
anwilkie: I added the directory to the docker-compose.yml file, so that the docker compose command uses the new file location.

PR #335, opened 2024-08-01
recreate dockerfile branch with signed commits

anwilkie:
# What and why?
This PR moves the dockerfile to the _infra/docker directory and adds signed commits.

In addition, the certifi dependency has been updated.

# How to test?
Check GA build.

# Jira
[Card](https://jira.ons.gov.uk/browse/RAS-313)


arroyoAle (approved):
looks good, build works, and docker compose works as expected

Pipfile.lock
matthew-robinson-ons, line 1: Was this supposed to be updated?

matthew-robinson-ons (changes requested):
There are a lot of changes in the pipfile.lock which i can't see documented. Will be hard to work out what has gone wrong if something breaks.

Pipfile.lock
anwilkie: Yes, there were vulnerabilities that required dependency updates.

anwilkie: I ran commands to update the specific packages in question. I don't know why the other packages got updated; I think that's just how pipenv works sometimes. If you're concerned about those updates, feel free to deploy the PR to your environment and test it out.

matthew-robinson-ons: You can use `pipenv upgrade` to update a specific package and use pipenv graph to display the dependencies so you can target the correct package and subpackages. Your card doesn't have any runtime changes but by updating all those packages it now does and the testing for it is a lot more in-depth than 'check GA build'. Aiming to make the smallest number of necessary changes seems t]like the better approach to me than just checking in a load of stuff that we don't know the effects of.

matthew-robinson-ons (approved):
Looks good and dependencies/vulnerabilites updated

PR #353, opened 2025-06-03
Introduce registry-instrument resource

pricem14pc:
# What and why?

Introduction of the new registry instrument resource.

>_Due to the complications detailed under [OPTION 1](https://github.com/ONSdigital/ras-rm-documentation/tree/main/spike/cir-integration#option-1) in the original design proposal, this solution does NOT address the already complex and redundant RASRM survey scoped collection instrument relationships. Indeed, this solution to introduce CIR deliberately sidesteps that data model, and introduces selected CIR resources, named `registry-instrument`, and keyed on RASRM UUID references, and explicitly operated on (CRUD) by its own API._

>_The consequence of this is loose data integrity, where the foreign and primary keys of the integer ID sequences for surveys, exercises and instruments are not used (as these are provided solely by survey scoped association tables). The new CIR instruments will be referenced by the UUID of a collection exercise and its explicit `formtype` classifier._

>_The CRUD operations required on these new resources will be done from within the context of the collection-exercise object in Rops. As such, the keys by which these CIR objects will be referenced will be within that scope and available from within the CIR pages._

>_Extending the existing API to include the selected CIR objects when returning the RM collection instruments is out of scope for this card (although that could be introduced at a later date)._

>_This means that additional/explicit HTTP requests will be required to request the selected CIR objects, and to interpolate the registry instruments with our collection instruments when building the pages. However, this will be straight forward, and requests will be non-complex, infrequent and fairly light._

>_This is intended to be an interim design, and would be completely replaced if/when the more appropriate [OPTION 2](https://github.com/ONSdigital/ras-rm-documentation/tree/main/spike/cir-integration#option-2) is delivered. There is no point in re-engineering the current poor data model to accommodate CIR._

- `RegistryInstrument` model and db tables
- Alembic migration for existing environments
- Flask view for `registry_instrument`
- Controller for `registry_instrument`
- API design for `RegistryInstrument` objects
- API implementation to `PUT` (create or update) registry instrument for a collection exercise ID
- API implementation to `GET` registry instruments by collection exercise ID
- API implementation to `GET` a registry instrument by collection exercise ID and form type
- API implementation to `DELETE` a registry instrument by collection exercise ID and form type

When working with the new APIs in the context of CIR features, the pages that will be making requests will have all the necessary request data in scope.

For example, the `/response_operations_ui/views/collection_exercise.py` has the user selected instrument data (including `form_type` and `instrument_id` values) in scope in the `linked_eq_ci` object, and the `collection_exercise_id` is in scope as `ce_details["collection_exercise"]["id"]`.

# How to test?

- build and run the tests locally
- deploy to your namespace
- use the API or run the SQL query below to insert or update selected CIR versions into the `registry_instrument` table
- use the API or run the SQL query below to retrieve selected CIR versions from the `registry_instrument` table

# API requests

> _Save a selected registry instrument for the given exercise UUID._

[OpenAPI spec](https://github.com/ONSdigital/ras-collection-instrument/blob/introduce-cir-instrument-resource/openapi.yaml#L432-L466)

```
curl --request PUT --location 'http://localhost:8002/collection-instrument-api/1.0.2/registry-instrument/exercise-id/59bc62d0-06a7-41ac-bd4a-6a7026ad94cd' \
--header 'Content-Type: application/json' \
--header 'Authorization: Basic YWRtaW46c2VjcmV0' \
--data '{
    "ci_version": 1,
    "classifier_type": "form_type",
    "classifier_value": "0007",
    "exercise_id": "3ff59b73-7f15-406f-9e4d-7f00b41e85ce",
    "guid": "c046861a-0df7-443a-a963-d9aa3bddf328",
    "instrument_id": "efc3ddd7-3e79-4c6b-a8f8-1fa184cdd06b",
    "published_at": "2025-12-31T00:00:00",
    "survey_id": "0b1f8376-28e9-4884-bea5-acf9d709464e"
}'
```

...should yield an HTTP `201` or `200` that the selected registry instrument version was created or updated for the collection exercise...

```
201 Successfully saved registry instrument
200 Successfully saved registry instrument
```

> _Get a list of selected registry instruments for the given exercise UUID._

[OpenAPI spec](https://github.com/ONSdigital/ras-collection-instrument/blob/introduce-cir-instrument-resource/openapi.yaml#L467-L490)

```
curl --location 'http://localhost:8002/collection-instrument-api/1.0.2/registry-instrument/exercise-id/3ff59b73-7f15-406f-9e4d-7f00b41e85ce' \
--header 'Authorization: Basic YWRtaW46c2VjcmV0'
```

...should yield a list of registry instrument versions selected for form types for the collection exercise...

```
[
        {
            "ci_version": 1,
            "classifier_type": "form_type",
            "classifier_value": "0001",
            "exercise_id": "3ff59b73-7f15-406f-9e4d-7f00b41e85ce",
            "guid": "c046861a-0df7-443a-a963-d9aa3bddf328",
            "instrument_id": "dfc3ddd7-3e79-4c6b-a8f8-1fa184cdd06b",
            "published_at": "2025-12-31T00:00:00",
            "survey_id": "0b1f8376-28e9-4884-bea5-acf9d709464e"
        },
        {
            "ci_version": 3,
            "classifier_type": "form_type",
            "classifier_value": "0002",
            "exercise_id": "3ff59b73-7f15-406f-9e4d-7f00b41e85ce",
            "guid": "ac3c5a3a-2ebb-47dc-9727-22c473086a82",
            "instrument_id": "77c1e406-716a-488d-bfc9-b5b988fbaccf",
            "published_at": "2025-12-31T12:00:00",
            "survey_id": "0b1f8376-28e9-4884-bea5-acf9d709464e"
        }
]
```

> _Get a selected registry instrument for the given exercise UUID and form type classifier value._

[OpenAPI spec](https://github.com/ONSdigital/ras-collection-instrument/blob/introduce-cir-instrument-resource/openapi.yaml#L492-L523)

```
curl --location 'http://localhost:8002/collection-instrument-api/1.0.2/registry-instrument/exercise-id/3ff59b73-7f15-406f-9e4d-7f00b41e85ce/formtype/0001' \
--header 'Authorization: Basic YWRtaW46c2VjcmV0'
```

...should yield the registry instrument version selected for the form type for the collection exercise...

```
{
    "ci_version": 1,
    "classifier_type": "form_type",
    "classifier_value": "0001",
    "exercise_id": "3ff59b73-7f15-406f-9e4d-7f00b41e85ce",
    "guid": "c046861a-0df7-443a-a963-d9aa3bddf328",
    "instrument_id": "dfc3ddd7-3e79-4c6b-a8f8-1fa184cdd06b",
    "published_at": "2025-12-31T00:00:00",
    "survey_id": "0b1f8376-28e9-4884-bea5-acf9d709464e"
}
```

> _Delete a selected registry instrument for the given exercise UUI and form type classifier value._

[Open API spec](https://github.com/ONSdigital/ras-collection-instrument/blob/introduce-cir-instrument-resource/openapi.yaml#L524-L554)

```
curl --request DELETE --location 'http://localhost:8002/collection-instrument-api/1.0.2/registry-instrument/exercise-id/27af467d-4dfe-4d23-b43c-2b8f1b52f25e/formtype/0004' \
--header 'Authorization: Basic YWRtaW46c2VjcmV0'
```
...should yield...
```
200 Successfully deleted registry instrument
```

# SQL Statements

You can also shove data into the DB with the following statements (`survey_id` and `collection_exercise_id` are the first two values respectively). You might want these to align with your own test data depending on what you're doing)

```
INSERT into ras_ci.registry_instrument
VALUES('0b1f8376-28e9-4884-bea5-acf9d709464e',
	  '3ff59b73-7f15-406f-9e4d-7f00b41e85ce',
	  '4952b0f2-838f-4902-a441-b41a4cd9d3bc',
	  'form_type',
	  '0001',
	  1,
	  'c046861a-0df7-443a-a963-d9aa3bddf328',
          '2025-12-31T00:00:00');

INSERT into ras_ci.registry_instrument
VALUES('0b1f8376-28e9-4884-bea5-acf9d709464e',
	  '3ff59b73-7f15-406f-9e4d-7f00b41e85ce',
	  '77c1e406-716a-488d-bfc9-b5b988fbaccf',
	  'form_type',
	  '0002',
	  1,
	  'ac3c5a3a-2ebb-47dc-9727-22c473086a82',
          '2025-12-31T00:00:00');
```

For reference, once selected registry instruments exist, their relationship with the survey scoped collection instruments can be seen with this reference query

```
SELECT
  a.survey_id,
  b.exercise_id,
  c.instrument_id,
  c.type,
  c.classifiers,
  e.classifier_value,
  e.ci_version,
  e.guid
 
FROM ras_ci.survey a
    FULL OUTER JOIN ras_ci.instrument c
	  ON a.id = c.survey_id
    FULL OUTER JOIN ras_ci.instrument_exercise d
	  ON  c.id = d.instrument_id
    FULL OUTER JOIN ras_ci.exercise b
	  ON b.id = d.exercise_id
    FULL OUTER JOIN ras_ci.registry_instrument e
	  ON b.exercise_id = e.exercise_id

WHERE
  a.survey_id = '0b1f8376-28e9-4884-bea5-acf9d709464e'
  AND b.exercise_id = '59bc62d0-06a7-41ac-bd4a-6a7026ad94cd'
  AND e.classifier_type = 'form_type'
  AND e.classifier_value = c.classifiers->>'form_type';
```

# Jira

https://jira.ons.gov.uk/browse/RAS-1603

### DIFF
--- a/application/controllers/registry_instrument.py
+++ b/application/controllers/registry_instrument.py
@@ -15,7 +15,7 @@ log = structlog.wrap_logger(logging.getLogger(__name__))
 
 class RegistryInstrument(object):
     @with_db_session
-    def get_registry_instruments_by_exercise_id(self, exercise_id, session=None):
+    def get_by_exercise_id(self, exercise_id, session=None):
         """
         Retrieves a list of selected CIR instruments for the given collection exercise
 
@@ -33,7 +33,7 @@ class RegistryInstrument(object):
         return response
 
     @with_db_session
-    def get_registry_instrument_by_exercise_id_and_formtype(self, exercise_id, form_type, session=None):
+    def get_by_exercise_id_and_formtype(self, exercise_id, form_type, session=None):
         """
         Retrieves the selected CIR instrument for the given collection exercise and form type
 
@@ -53,7 +53,7 @@ class RegistryInstrument(object):
         return None
 
     @with_db_session
-    def save_registry_instrument_for_exercise_id_and_formtype(
+    def save_for_exercise_id_and_formtype(
         self, survey_id, exercise_id, instrument_id, form_type, ci_version, published_at, guid, session=None
     ):
         """
@@ -65,7 +65,7 @@ class RegistryInstrument(object):
         :return: True if successful, and is_new indicating if a new object was created
         """
 
-        registry_instrument, is_new = self._find_or_create_registry_instrument(
+        registry_instrument, is_new = self._find_or_create(
             survey_id, exercise_id, instrument_id, form_type, ci_version, published_at, guid, session
         )
 
@@ -73,7 +73,7 @@ class RegistryInstrument(object):
         return True, is_new
 
     @with_db_session
-    def delete_registry_instrument_by_exercise_id_and_formtype(self, exercise_id, form_type, session=None):
+    def delete_by_exercise_id_and_formtype(self, exercise_id, form_type, session=None):
         """
         Delete the selected CIR instrument for the given collection exercise and form type
 
@@ -95,9 +95,7 @@ class RegistryInstrument(object):
             return False
 
     @staticmethod
-    def _find_or_create_registry_instrument(
-        survey_id, exercise_id, instrument_id, form_type, ci_version, published_at, guid, session
-    ):
+    def _find_or_create(survey_id, exercise_id, instrument_id, form_type, ci_version, published_at, guid, session):
         """
         Retrieves an existing selected RegistryInstrumentModel object from the db if it exists,
         or creates a new RegistryInstrumentModel object if it doesn't exist
--- a/application/controllers/registry_instrument.py
+++ b/application/controllers/registry_instrument.py
@@ -111,7 +111,13 @@ class RegistryInstrument(object):
             exercise_id, form_type, session
         ).first()
 
-        if not registry_instrument:
+        if registry_instrument:
+            log.info("Registry instrument found, updating", exercise_id=exercise_id, form_type=form_type)
+            registry_instrument.ci_version = ci_version
+            registry_instrument.guid = guid
+            registry_instrument.published_at = published_at
+            return registry_instrument, False
+        else:
             log.info("Registry instrument NOT found, creating", exercise_id=exercise_id, form_type=form_type)
             registry_instrument = RegistryInstrumentModel()
             registry_instrument.survey_id = survey_id
@@ -122,12 +128,4 @@ class RegistryInstrument(object):
             registry_instrument.ci_version = ci_version
             registry_instrument.guid = guid
             registry_instrument.published_at = published_at
-            is_new = True
-        else:
-            log.info("Registry instrument found, updating", exercise_id=exercise_id, form_type=form_type)
-            registry_instrument.ci_version = ci_version
-            registry_instrument.guid = guid
-            registry_instrument.published_at = published_at
-            is_new = False
-
-        return registry_instrument, is_new
+            return registry_instrument, True
--- a/application/controllers/registry_instrument.py
+++ b/application/controllers/registry_instrument.py
@@ -8,6 +8,7 @@ from application.controllers.sql_queries import (
     query_registry_instrument_by_exercise_id_and_formtype,
     query_registry_instruments_by_exercise_id,
 )
+from application.models.models import RegistryInstrumentModel
 
 log = structlog.wrap_logger(logging.getLogger(__name__))
 
@@ -34,12 +35,12 @@ class RegistryInstrument(object):
     @with_db_session
     def get_registry_instrument_by_exercise_id_and_formtype(self, exercise_id, form_type, session=None):
         """
-        Retrieves a selected CIR instrument for the given collection exercise and form type
+        Retrieves the selected CIR instrument for the given collection exercise and form type
 
         :param exercise_id: An exercise id (UUID)
         :param form_type: The form type (e.g. "0001")
         :param session: database session
-        :return: A RegistryInstrumentModel object
+        :return: A RegistryInstrumentModel object, or None if not found
         """
         log.info("Retrieving a selected CIR instrument", exercise_id=exercise_id, form_type=form_type)
         validate_uuid(exercise_id)
@@ -50,3 +51,68 @@ class RegistryInstrument(object):
         if registry_instrument is not None:
             return registry_instrument.to_dict()
         return None
+
+    @with_db_session
+    def save_registry_instrument_for_exercise_id_and_formtype(
+        self, survey_id, exercise_id, instrument_id, form_type, ci_version, published_at, guid, session=None
+    ):
+        """
+        Saves a selected CIR instrument for the given collection exercise and form type
+
+        :param exercise_id: An exercise id (UUID)
+        :param form_type: The form type (e.g. "0001")
+        :param session: database session
+        :return: True if successful, and is_new indicating if a new object was created
+        """
+
+        registry_instrument, is_new = self._find_or_create_registry_instrument(
+            survey_id, exercise_id, instrument_id, form_type, ci_version, published_at, guid, session
+        )
+
+        if registry_instrument is not None:
+            if not is_new:
+                registry_instrument.ci_version = ci_version
+                registry_instrument.guid = guid
+                registry_instrument.published_at = published_at
+            session.add(registry_instrument)
+            return True, is_new
+        else:
+            return False, None
+
+    @staticmethod
+    def _find_or_create_registry_instrument(
+        survey_id, exercise_id, instrument_id, form_type, ci_version, published_at, guid, session
+    ):
+        """
+        Retrieves an existing selected RegistryInstrumentModel object from the db if it exists,
+        or creates a new RegistryInstrumentModel object if it doesn't exist
+
+        :param exercise_id: An exercise id (UUID)
+        :param form_type: The form type (e.g. "0001")
+        :param session: database session
+        :return: RegistryInstrumentModel object
+        """
+
+        validate_uuid(exercise_id)
+        registry_instrument = query_registry_instrument_by_exercise_id_and_formtype(
+            exercise_id, form_type, session
+        ).first()
+
+        if not registry_instrument:
+            log.info("Registry instrument NOT found, creating new object", exercise_id=exercise_id, form_type=form_type)
+            registry_instrument = RegistryInstrumentModel(
+                survey_id=survey_id,
+                exercise_id=exercise_id,
+                instrument_id=instrument_id,
+                classifier_type="form_type",
+                classifier_value=form_type,
+                ci_version=ci_version,
+                guid=guid,
+                published_at=published_at,
+            )
+            is_new = True
+        else:
+            log.info("Registry instrument found", exercise_id=exercise_id, form_type=form_type)
+            is_new = False
+
+        return registry_instrument, is_new
--- a/application/controllers/registry_instrument.py
+++ b/application/controllers/registry_instrument.py
@@ -88,11 +88,11 @@ class RegistryInstrument(object):
             exercise_id, form_type, session
         ).first()
 
-        if registry_instrument is not None:
-            session.delete(registry_instrument)
-            return True
-        else:
+        if registry_instrument is None:
             return False
+
+        session.delete(registry_instrument)
+        return True
 
     @staticmethod
     def _find_or_create(survey_id, exercise_id, instrument_id, form_type, ci_version, published_at, guid, session):
--- a/application/controllers/registry_instrument.py
+++ b/application/controllers/registry_instrument.py
@@ -14,7 +14,7 @@ log = structlog.wrap_logger(logging.getLogger(__name__))
 
 class RegistryInstrument(object):
     @with_db_session
-    def get_registry_instruments_by_exercise_id(exercise_id, session=None):
+    def get_registry_instruments_by_exercise_id(self, exercise_id, session=None):
         """
         Retrieves a list of selected CIR instruments for the given collection exercise
 
@@ -28,7 +28,7 @@ class RegistryInstrument(object):
         return registry_instruments
 
     @with_db_session
-    def get_registry_instrument_by_exercise_id_and_formtype(exercise_id, form_type, session=None):
+    def get_registry_instrument_by_exercise_id_and_formtype(self, exercise_id, form_type, session=None):
         """
         Retrieves a selected CIR instrument for the given collection exercise and form type
 
--- a/application/controllers/registry_instrument.py
+++ b/application/controllers/registry_instrument.py
@@ -14,7 +14,7 @@ log = structlog.wrap_logger(logging.getLogger(__name__))
 
 class RegistryInstrument(object):
     @with_db_session
-    def get_registry_instruments_by_exercise_id(exercise_id, session):
+    def get_registry_instruments_by_exercise_id(exercise_id, session=None):
         """
         Retrieves a list of selected CIR instruments for the given collection exercise
 
@@ -24,7 +24,7 @@ class RegistryInstrument(object):
         """
         log.info("Retrieving list of selected CIR instruments", exercise_id=exercise_id)
         validate_uuid(exercise_id)
-        registry_instruments = query_registry_instruments_by_exercise_id(exercise_id, session=None)
+        registry_instruments = query_registry_instruments_by_exercise_id(exercise_id, session)
         return registry_instruments
 
     @with_db_session
--- a/application/controllers/registry_instrument.py
+++ b/application/controllers/registry_instrument.py
@@ -20,12 +20,16 @@ class RegistryInstrument(object):
 
         :param exercise_id: An exercise id (UUID)
         :param session: database session
-        :return: list of RegistryInstrumentModel objects
+        :return: a list of RegistryInstrumentModel dictionaries
         """
         log.info("Retrieving list of selected CIR instruments", exercise_id=exercise_id)
         validate_uuid(exercise_id)
         registry_instruments = query_registry_instruments_by_exercise_id(exercise_id, session)
-        return registry_instruments
+
+        response = []
+        for registry_instrument in registry_instruments:
+            response.append(registry_instrument.to_dict())
+        return response
 
     @with_db_session
     def get_registry_instrument_by_exercise_id_and_formtype(self, exercise_id, form_type, session=None):
@@ -39,5 +43,10 @@ class RegistryInstrument(object):
         """
         log.info("Retrieving a selected CIR instrument", exercise_id=exercise_id, form_type=form_type)
         validate_uuid(exercise_id)
-        registry_instrument = query_registry_instrument_by_exercise_id_and_formtype(exercise_id, form_type, session)
-        return registry_instrument
+        registry_instrument = query_registry_instrument_by_exercise_id_and_formtype(
+            exercise_id, form_type, session
+        ).first()
+
+        if registry_instrument is not None:
+            return registry_instrument.to_dict()
+        return None
--- a/application/controllers/registry_instrument.py
+++ b/application/controllers/registry_instrument.py
@@ -57,7 +57,7 @@ class RegistryInstrument(object):
         self, survey_id, exercise_id, instrument_id, form_type, ci_version, published_at, guid, session=None
     ):
         """
-        Saves a selected CIR instrument for the given collection exercise and form type
+        Save a selected CIR instrument for the given collection exercise and form type
 
         :param exercise_id: An exercise id (UUID)
         :param form_type: The form type (e.g. "0001")
@@ -78,6 +78,28 @@ class RegistryInstrument(object):
             return True, is_new
         else:
             return False, None
+
+    @with_db_session
+    def delete_registry_instrument_by_exercise_id_and_formtype(self, exercise_id, form_type, session=None):
+        """
+        Delete the selected CIR instrument for the given collection exercise and form type
+
+        :param exercise_id: An exercise id (UUID)
+        :param form_type: The form type (e.g. "0001")
+        :param session: database session
+        :return: A boolean indicating success
+        """
+        log.info("Deleting a selected CIR instrument", exercise_id=exercise_id, form_type=form_type)
+        validate_uuid(exercise_id)
+        registry_instrument = query_registry_instrument_by_exercise_id_and_formtype(
+            exercise_id, form_type, session
+        ).first()
+
+        if registry_instrument is not None:
+            session.delete(registry_instrument)
+            return True
+        else:
+            return False
 
     @staticmethod
     def _find_or_create_registry_instrument(
--- a/application/controllers/registry_instrument.py
+++ b/application/controllers/registry_instrument.py
@@ -122,16 +122,15 @@ class RegistryInstrument(object):
 
         if not registry_instrument:
             log.info("Registry instrument NOT found, creating new object", exercise_id=exercise_id, form_type=form_type)
-            registry_instrument = RegistryInstrumentModel(
-                survey_id=survey_id,
-                exercise_id=exercise_id,
-                instrument_id=instrument_id,
-                classifier_type="form_type",
-                classifier_value=form_type,
-                ci_version=ci_version,
-                guid=guid,
-                published_at=published_at,
-            )
+            registry_instrument = RegistryInstrumentModel()
+            registry_instrument.survey_id = survey_id
+            registry_instrument.exercise_id = exercise_id
+            registry_instrument.instrument_id = instrument_id
+            registry_instrument.classifier_type = "form_type"
+            registry_instrument.classifier_value = form_type
+            registry_instrument.ci_version = ci_version
+            registry_instrument.guid = guid
+            registry_instrument.published_at = published_at
             is_new = True
         else:
             log.info("Registry instrument found", exercise_id=exercise_id, form_type=form_type)
--- a/application/controllers/registry_instrument.py
+++ b/application/controllers/registry_instrument.py
@@ -69,15 +69,8 @@ class RegistryInstrument(object):
             survey_id, exercise_id, instrument_id, form_type, ci_version, published_at, guid, session
         )
 
-        if registry_instrument is not None:
-            if not is_new:
-                registry_instrument.ci_version = ci_version
-                registry_instrument.guid = guid
-                registry_instrument.published_at = published_at
-            session.add(registry_instrument)
-            return True, is_new
-        else:
-            return False, None
+        session.add(registry_instrument)
+        return True, is_new
 
     @with_db_session
     def delete_registry_instrument_by_exercise_id_and_formtype(self, exercise_id, form_type, session=None):
@@ -121,7 +114,7 @@ class RegistryInstrument(object):
         ).first()
 
         if not registry_instrument:
-            log.info("Registry instrument NOT found, creating new object", exercise_id=exercise_id, form_type=form_type)
+            log.info("Registry instrument NOT found, creating", exercise_id=exercise_id, form_type=form_type)
             registry_instrument = RegistryInstrumentModel()
             registry_instrument.survey_id = survey_id
             registry_instrument.exercise_id = exercise_id
@@ -133,7 +126,10 @@ class RegistryInstrument(object):
             registry_instrument.published_at = published_at
             is_new = True
         else:
-            log.info("Registry instrument found", exercise_id=exercise_id, form_type=form_type)
+            log.info("Registry instrument found, updating", exercise_id=exercise_id, form_type=form_type)
+            registry_instrument.ci_version = ci_version
+            registry_instrument.guid = guid
+            registry_instrument.published_at = published_at
             is_new = False
 
         return registry_instrument, is_new
--- a/application/controllers/registry_instrument.py
+++ b/application/controllers/registry_instrument.py
@@ -3,6 +3,7 @@ import logging
 import structlog
 
 from application.controllers.helper import validate_uuid
+from application.controllers.session_decorator import with_db_session
 from application.controllers.sql_queries import (
     query_registry_instrument_by_exercise_id_and_formtype,
     query_registry_instruments_by_exercise_id,
@@ -12,7 +13,7 @@ log = structlog.wrap_logger(logging.getLogger(__name__))
 
 
 class RegistryInstrument(object):
-    @staticmethod
+    @with_db_session
     def get_registry_instruments_by_exercise_id(exercise_id, session):
         """
         Retrieves a list of selected CIR instruments for the given collection exercise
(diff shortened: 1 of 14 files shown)


application/models/models.py
```diff
@@ -189,3 +189,53 @@ def __init__(self, instrument_id=None, file_name=None, length=None, data=None):
...
+    * In RASRM the survey_id is a UUID, however in the context of the CIR object being returned, their survey_id is
+      what we know of as the three digit survey_ref e.g. "139". This table will hold OUR survey_id as a UUID and
+      not the three digit survey_ref returned by the CIR API.
+    """
+
+    __tablename__ = "registry_instrument"
+
+    # survey_id = Column(UUID, ForeignKey("survey.survey_id"))
+    # exercise_id = Column(UUID, ForeignKey("exercise.exercise_id"), primary_key=True)
+    # instrument_id = Column(UUID, ForeignKey("instrument.instrument_id"))
```
pricem14pc, line 212: These constraints will be added in once the stubs are removed, so it's easier to shovel some data in to the DB without having to have all the survey and collection exercise records in place.

application/views/registry_instrument_view.py
```diff
@@ -0,0 +1,77 @@
...
+registry_instrument_view = Blueprint("registry_instrument_view", __name__)
+
+
+@registry_instrument_view.before_request
+@auth.login_required
+def before_registry_instrument_view():
+    pass
+
+
+@registry_instrument_view.route("/registry-instrument/exercise-id/<exercise_id>", methods=["GET"])
```
pricem14pc, line 20: Happy to debate the appropriateness of the name of this new resource. However, while implementing and trying different names I settled on `registry-instrument` and went through many different names.

Although similar to the existing named RASRM `collection-instrument`, there is no good way to name it while showing it's referring to a distinct external object.

application/views/registry_instrument_view.py
```diff
@@ -0,0 +1,39 @@
...
+def get_registry_instruments(exercise_id):
+    registry_instruments = RegistryInstrument().get_registry_instruments_by_exercise_id(exercise_id)
+
+    if registry_instruments:
+        return make_response(jsonify(registry_instruments), 200)
+
+    return make_response("Not Found", 404)
```
pricem14pc, line 27: As per the comment in the API doc, I wonder if this would be better to return a `200` and an empty list `[ ]`. Its a valid scenario that there might not yet be any selected registry instruments for the given collection exercise. Returning a `404` feels heavy handed and makes the Rops usage a bit more complicated.

Comments welcomed?

application/models/models.py
```diff
@@ -189,3 +189,65 @@ def __init__(self, instrument_id=None, file_name=None, length=None, data=None):
...
+    def __init__(
+        self,
+        survey_id=None,
+        exercise_id=None,
+        instrument_id=None,
+        classifier_type=None,
+        classifier_value=None,
+        ci_version=None,
+        guid=None,
+        published_at=None,
```
pricem14pc, line 231: I'm thinking about introducing a DTO class to represent these as I'm handing them round quite a lot. A bit smelly.

application/views/registry_instrument_view.py
```diff
@@ -0,0 +1,153 @@
...
+
+    if success:
+        return make_response(
+            "Successfully saved registry instrument",
+            HTTPStatus.CREATED if is_new else HTTPStatus.OK,
+        )
+    else:
+        return make_response(
+            "Registry instrument failed to save successfully",
+            HTTPStatus.INTERNAL_SERVER_ERROR,
```
pricem14pc, line 152: This should never be possible. If it gets here something really unexpected has happened hence the `500`

application/views/registry_instrument_view.py
```diff
@@ -0,0 +1,176 @@
...
+        "ci_version",
+        "guid",
+        "published_at",
+    }
+
+    payload_keys = set(payload.keys())
+
+    if payload_keys != expected_keys:
+        return make_response(
+            f"Invalid payload keys. Expected: {expected_keys}, Included: {payload_keys}",
```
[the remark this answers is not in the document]
pricem14pc: I will look at an alternative way to validating these, however the 'user provided value' is misleading as the request is only ever formed and made by our Rops system. We also subsequently validate the keys and values in the next part of the flow. As such, I believe this alert is spurious.

application/views/registry_instrument_view.py
matthew-robinson-ons: I can see an argument for both. When a new collection exercise is setup it would be expected and valid that no registry instruments are selected and that does feel heavy handed to be using a 404. Though is there a gap where we could return a 200 and an empty list unintentionally? I lean towards the 404 but i'm not sure how much more complicated Rops would be to implement then.

application/views/registry_instrument_view.py
matthew-robinson-ons: I think this name is appropriate, when reviewing a previous card i was struggling to think of a name for these resources and this seems an appropriate name for them.

application/models/models.py
```diff
@@ -189,3 +189,53 @@ def __init__(self, instrument_id=None, file_name=None, length=None, data=None):
         self.instrument_id = instrument_id
         self.file_name = file_name
         self.len = length
+
+
+class RegistryInstrumentModel(Base):
```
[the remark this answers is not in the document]
matthew-robinson-ons: Commented below but i think this is an appropriate name for the resource.

openapi.yaml
```diff
@@ -426,6 +428,76 @@ paths:
...
+        '200':
+          description: Returns a list of selected registry instruments for the given exercise UUID.
+          content:
+            application/json:
+              schema:
+                type: array
+                items:
+                  $ref: '#/components/schemas/RegistryInstrument'
+        '404':
+          description: No registry instruments found for the given exercise ID
```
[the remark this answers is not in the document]
pricem14pc: Off-line decision to return an empty list `[ ]` and 200 status if no registry instruments have been selected for the given exercise UUID. 404 will be returned if the exercise UUID does not exist.

application/views/registry_instrument_view.py
pricem14pc: Done

openapi.yaml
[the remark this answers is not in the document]
pricem14pc: Done

tests/controllers/test_registry_instrument.py
```diff
@@ -0,0 +1,196 @@
...
+from unittest.mock import MagicMock, patch
+
+from application.controllers.registry_instrument import RegistryInstrument
+from application.exceptions import RasError
+
+
+class TestRegistryInstrumentController(TestCase):
+
+    @patch("application.controllers.registry_instrument.query_registry_instruments_by_exercise_id")
+    def test_get_registry_instruments_by_exercise_id_returns_list_of_dicts(self, mock_query):
```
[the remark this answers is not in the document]
LJBabbage: We can shrink this one down to make it more digestible we don't need to do this
``` mock_query.assert_called_once() ``` there is no lop and it's a one to one, it's just noise asserting it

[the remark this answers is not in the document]
Comment (LJBabbage) [reply tests/controllers/test_registry_instrument.py:11]: Equally having 2 objects and asserting both call to_dict once is just a bit over kill. I wouldn't even bother calling to them at all, the result tells you that.

Comment (LJBabbage) [inline tests/controllers/test_registry_instrument.py:11]: mock_query is a touch lazy, nice to have a true reflection of what it is. It's not so bad here, but when you have multiple patches it's difficult to keep track

File: tests/controllers/test_registry_instrument.py [hunk 10]
```diff
@@ -0,0 +1,196 @@
...
+        controller = RegistryInstrument()
+        exercise_id = "3ff59b73-7f15-406f-9e4d-7f00b41e85ce"
+        formtype = "0001"
+
+        result = controller.get_registry_instrument_by_exercise_id_and_formtype.__wrapped__(
+            controller, exercise_id, formtype, session
+        )
+
+        mock_query.assert_called_once()
+        mock_registry_instrument.to_dict.assert_called_once()
```
LJBabbage, line 62: Same again, and I won't mention it again, I don't think this is necessary, with the result as well. Belt, braces and hands in pockets

tests/controllers/test_registry_instrument.py
```diff
@@ -0,0 +1,196 @@
...
+        form_type = "0002"
+        exercise_id = "3ff59b73-7f15-406f-9e4d-7f00b41e85ce"
+        guid = "678a583a-87f6-4c57-9f5b-12c5ced30c1e"
+        published_at = "2028-01-30T12:00:00"
+        survey_id = "0b1f8376-28e9-4884-bea5-acf9d709464e"
+        instrument_id = "77c1e406-716a-488d-bfc9-b5b988fbaccf"
+
+        controller = RegistryInstrument()
+
+        with patch(
```
LJBabbage, line 102: Mid test patch? why not a decoator

tests/controllers/test_registry_instrument.py
```diff
@@ -0,0 +1,196 @@
...
+            "application.controllers.registry_instrument.RegistryInstrumentModel"
+        ) as mock_registry_instrument_model:
+            result = controller.save_registry_instrument_for_exercise_id_and_formtype.__wrapped__(
+                controller, survey_id, exercise_id, instrument_id, form_type, ci_version, published_at, guid, session
+            )
+
+        # Assert that a new instrument was created and saved
+        mock_query.assert_called_once()
+        mock_registry_instrument_model.assert_called_once()
+        self.assertEqual(mock_registry_instrument_model.return_value.ci_version, ci_version)
```
LJBabbage, line 112: Feels like this could be done a bit neater, i.e called to dict and then compare that dict to what you expect

tests/controllers/test_registry_instrument.py
```diff
@@ -0,0 +1,196 @@
...
+        controller = RegistryInstrument()
+        exercise_id = "invalid_uuid"
+
+        with self.assertRaises(RasError) as context:
+            controller.get_registry_instruments_by_exercise_id.__wrapped__(controller, exercise_id, session)
+
+        self.assertEqual(context.exception.errors[0], "Value is not a valid UUID (invalid_uuid)")
+
+    @patch("application.controllers.registry_instrument.query_registry_instrument_by_exercise_id_and_formtype")
+    def test_get_registry_instrument_by_exercise_id_and_formtype_returns_dict(self, mock_query):
```
LJBabbage, line 46: As before with the name, I don't include the expected return behaviour only the negative

tests/controllers/test_registry_instrument.py
```diff
@@ -0,0 +1,196 @@
...
+        result = controller.get_registry_instrument_by_exercise_id_and_formtype.__wrapped__(
+            controller, exercise_id, formtype, session
+        )
+
+        mock_query.assert_called_once()
+        mock_registry_instrument.to_dict.assert_called_once()
+        self.assertEqual(result, {"dummy_key": "dummy_value_1"})
+
+    @patch("application.controllers.registry_instrument.query_registry_instrument_by_exercise_id_and_formtype")
+    def test_get_registry_instrument_by_exercise_id_and_formtype_returns_none(self, mock_query):
```
LJBabbage, line 66: Im thinking this is a missing exercise_id call. returns_none doesn't really tell me much

tests/controllers/test_registry_instrument.py
```diff
@@ -0,0 +1,196 @@
...
+
+    @patch("application.controllers.registry_instrument.query_registry_instrument_by_exercise_id_and_formtype")
+    def test_save_new_registry_instrument_for_exercise_id_and_formtype(self, mock_query):
+        session = MagicMock()
+
+        # Mock a registry instrument as NOT in the database
+        mock_query.return_value.first.return_value = None
+
+        # Set up the validated data from the posted payload
+        ci_version = 99
```
LJBabbage, line 92: Could these be global and extracted?

tests/controllers/test_registry_instrument.py
```diff
@@ -0,0 +1,196 @@
...
+        mock_registry_instrument = MagicMock()
+        mock_registry_instrument.to_dict.return_value = {
+            "ci_version": 3,
+            "guid": "ac3c5a3a-2ebb-47dc-9727-22c473086a82",
+            "published_at": "2025-12-31T12:00:00",
+        }
+        mock_query.return_value.first.return_value = mock_registry_instrument
+
+        # Set up the validated data from the posted payload
+        ci_version = 99
```
LJBabbage, line 137: Ok, so to answer my own question, they can be extracted

tests/controllers/test_registry_instrument.py
```diff
@@ -0,0 +1,196 @@
...
+
+        controller = RegistryInstrument()
+
+        result = controller.save_registry_instrument_for_exercise_id_and_formtype.__wrapped__(
+            controller, None, exercise_id, None, form_type, ci_version, published_at, guid, session
+        )
+
+        # Assert that the existing instrument was updated and saved
+        mock_query.assert_called_once()
+        self.assertEqual(result, (True, False))
```
LJBabbage, line 151: As before possible could be neater and a single assert

tests/views/test_registry_instrument_view.py
```diff
@@ -0,0 +1,206 @@
...
+form_type_exists = "0001"
+form_type_does_not_exist = "9999"
+
+
+class TestRegistryInstrumentView(TestClient):
+    """
+    Registry Instrument view unit tests
+    """
+
+    def test_get_registry_instrument_returns_200_and_object(self):
```
LJBabbage, line 21: Probably best to stick to doing patches the same way

tests/views/test_registry_instrument_view.py
```diff
@@ -0,0 +1,206 @@
...
+            "RegistryInstrument.get_registry_instrument_by_exercise_id_and_formtype"
+        ) as mock_get_registry_instrument_by_exercise_id_and_formtype:
+            with open(Path(__file__).parent.parent / "test_data" / "registry_instrument.json") as f:
+                mock_get_registry_instrument_by_exercise_id_and_formtype.return_value = json.load(f)
+            response = self.client.get(
+                f"{api_root}/registry-instrument/exercise-id/{exercise_id}/formtype/{form_type_exists}",
+                headers=self.get_auth_headers(),
+            )
+            self.assertStatus(response, 200)
+            self.assertEqual(len(response.json), 8)  # The 8 fields of a registry instrument
```
LJBabbage, line 33: I think we need to assert the content not the size here

tests/views/test_registry_instrument_view.py
```diff
@@ -0,0 +1,206 @@
...
+            "application.views.registry_instrument_view."
+            "RegistryInstrument.get_registry_instrument_by_exercise_id_and_formtype"
+        ) as mock_get_registry_instrument_by_exercise_id_and_formtype:
+            mock_get_registry_instrument_by_exercise_id_and_formtype.return_value = None
+            response = self.client.get(
+                f"{api_root}/registry-instrument/exercise-id/{exercise_id}/formtype/{form_type_does_not_exist}",
+                headers=self.get_auth_headers(),
+            )
+            self.assertStatus(response, 404)
+            self.assertEqual(response.data.decode(), "Not Found")  # The 8 fields of a registry instrument
```
LJBabbage, line 50: Not sure on the comment? Feels like a copy paste error

tests/views/test_registry_instrument_view.py
```diff
@@ -0,0 +1,206 @@
...
+                f"{api_root}/registry-instrument/exercise-id/{exercise_id}/formtype/{form_type_exists}",
+                headers=self.get_auth_headers(),
+            )
+            self.assertEqual(response.data.decode(), "Successfully deleted registry instrument")
+            self.assertEqual(response.headers["Content-Type"], "text/html; charset=utf-8")
+            mock_delete_registry_instrument_by_exercise_id_and_formtype.assert_called_once_with(
+                exercise_id, form_type_exists
+            )
+
+    def test_unsuccessful_delete_registry_instrument_returns_404(self):
```
LJBabbage, line 101: Not a huge fan of this name, Im assuming it's because the registry_instrument isn't in the database, but the name is a bit verbose. It makes me think any failure is coming back 404. test_unsuccessful_delete_registry_instrument_missing_instrument

application/controllers/registry_instrument.py
```diff
@@ -0,0 +1,139 @@
...
+        :param form_type: The form type (e.g. "0001")
+        :param session: database session
+        :return: True if successful, and is_new indicating if a new object was created
+        """
+
+        registry_instrument, is_new = self._find_or_create_registry_instrument(
+            survey_id, exercise_id, instrument_id, form_type, ci_version, published_at, guid, session
+        )
+
+        if registry_instrument is not None:
```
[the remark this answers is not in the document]
pricem14pc: No, it should never be None. Belt and braces and probably a hangover from a previous iteration. Interestingly the resulting else (line 80) was the only line not covered by the tests as it's technically unreachable, so a good spot and I'll remove it.

application/controllers/registry_instrument.py
```diff
@@ -0,0 +1,139 @@
...
+            if not is_new:
+                registry_instrument.ci_version = ci_version
+                registry_instrument.guid = guid
+                registry_instrument.published_at = published_at
+            session.add(registry_instrument)
+            return True, is_new
+        else:
```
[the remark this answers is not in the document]
pricem14pc: See my comment above, you are right this is technically unreachable.

application/controllers/registry_instrument.py
```diff
@@ -0,0 +1,139 @@
...
+        :param session: database session
+        :return: True if successful, and is_new indicating if a new object was created
+        """
+
+        registry_instrument, is_new = self._find_or_create_registry_instrument(
+            survey_id, exercise_id, instrument_id, form_type, ci_version, published_at, guid, session
+        )
+
+        if registry_instrument is not None:
+            if not is_new:
```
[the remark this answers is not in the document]
pricem14pc: Do you mean setting the values for the new object inside the `_find_or_create_registry_instrument`? If so yes, this would be cleaner. But compromising that that method continues to do both the "find" or "create", but additionally sets the values? How does that sound?

application/controllers/registry_instrument.py
```diff
@@ -0,0 +1,139 @@
...
+        :param session: database session
+        :return: A boolean indicating success
+        """
+        log.info("Deleting a selected CIR instrument", exercise_id=exercise_id, form_type=form_type)
+        validate_uuid(exercise_id)
+        registry_instrument = query_registry_instrument_by_exercise_id_and_formtype(
+            exercise_id, form_type, session
+        ).first()
+
+        if registry_instrument is not None:
```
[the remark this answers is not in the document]
pricem14pc: I'm assuming you're advocating not doing the initial retrieval, and simply trying to delete it regardless of whether it exists or not?

application/views/registry_instrument_view.py
```diff
@@ -0,0 +1,117 @@
...
+        survey_id=payload["survey_id"],
+        exercise_id=payload["exercise_id"],
+        instrument_id=payload["instrument_id"],
+        form_type=payload["classifier_value"],
+        ci_version=payload["ci_version"],
+        published_at=payload["published_at"],
+        guid=payload["guid"],
+    )
+
+    if success:
```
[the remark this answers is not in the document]
pricem14pc: I think this has been covered by changes above?

The only way this exception will be truly exceptional, is if the front-end makes an initial request to check for an existing selected registry instrument, before then making a subsequent request to DELETE it. Otherwise we should deal with the possibility that the requested registry instrument to be deleted does NOT exist, and deal with that gracefully without raising an exception.

As mentioned, there will be scenarios where the user attempts to DELETE an already deleted registry instrument (i.e. keeps saving a "No CIR version chosen").

application/views/registry_instrument_view.py
```diff
@@ -0,0 +1,117 @@
...
+
+    # TODO: (low priority) check the exercise_id exists in the ras_ci.exercise table
+    #       and return 404 if it does not exist
+
+    return make_response(jsonify(registry_instruments), HTTPStatus.OK)
+
+
+@registry_instrument_view.route("/registry-instrument/exercise-id/<exercise_id>/formtype/<form_type>", methods=["GET"])
+def get_registry_instrument(exercise_id, form_type):
+    registry_instrument = RegistryInstrument().get_registry_instrument_by_exercise_id_and_formtype(
```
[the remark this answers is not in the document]
pricem14pc: At the moment I have them using the `@with_db_session` which they don't need as they're not transactional. So yes, I will change that. But let me know if are you suggesting making the various `get_` methods static in the controller?

application/views/registry_instrument_view.py
```diff
@@ -0,0 +1,117 @@
...
+@registry_instrument_view.route("/registry-instrument/exercise-id/<exercise_id>", methods=["GET"])
+def get_registry_instruments(exercise_id):
+    """
+    Returns an array of selected CIR instrument for the given collection exercise.
+
+    :param exercise_id: An exercise id (UUID)
+    """
+    registry_instruments = RegistryInstrument().get_registry_instruments_by_exercise_id(exercise_id)
+
+    # TODO: (low priority) check the exercise_id exists in the ras_ci.exercise table
```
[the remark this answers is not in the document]
pricem14pc: Done

application/views/registry_instrument_view.py
```diff
@@ -0,0 +1,117 @@
...
+             400 if the payload is invalid
+    """
+
+    try:
+        payload = request.get_json(force=True)
+        is_valid, error = validate_registry_instrument_payload(payload, exercise_id)
+        if not is_valid:
+            return make_response(error, HTTPStatus.BAD_REQUEST)
+    except BadRequest:
+        return make_response("Invalid JSON payload", HTTPStatus.BAD_REQUEST)
```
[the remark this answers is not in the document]
pricem14pc: But `request.get_json` throws a BadRequest. Hence this structure to catch malformed json.

application/controllers/registry_instrument.py
```diff
@@ -0,0 +1,139 @@
...
+
+        :param exercise_id: An exercise id (UUID)
+        :param session: database session
+        :return: a list of RegistryInstrumentModel dictionaries
+        """
+        log.info("Retrieving list of selected CIR instruments", exercise_id=exercise_id)
+        validate_uuid(exercise_id)
+        registry_instruments = query_registry_instruments_by_exercise_id(exercise_id, session)
+
+        response = []
```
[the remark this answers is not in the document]
LJBabbage: Yep, exactly what I was getting at

application/views/registry_instrument_view.py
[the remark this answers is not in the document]
LJBabbage: It should be ```    
try:
        payload = request.get_json(force=True)
    except BadRequest:
        return make_response("Invalid JSON payload", HTTPStatus.BAD_REQUEST)        
is_valid, error = validate_registry_instrument_payload(payload, exercise_id)
        if not is_valid:
            return make_response(error, HTTPStatus.BAD_REQUEST) ```
You don't need to have something that does throw an exception to be in the try. The only time you might do that is if you suppress the exception, but in this instance you don't as you are returning

tests/controllers/test_registry_instrument.py
LJBabbage: I generally name the mock after the patch, so if Im 10 lines down I know what it is rather than mock_query. So I do something like query_registry_instruments_by_exercise_id

pricem14pc: Oh sorry.. I thought you were saying I shouldn't be mocking the DB call at all :o) It was just a comment about the name. Sure, I'll update.

application/views/registry_instrument_view.py
```diff
@@ -0,0 +1,117 @@
...
+            HTTPStatus.INTERNAL_SERVER_ERROR,
+        )
+
+
+@registry_instrument_view.route(
+    "/registry-instrument/exercise-id/<exercise_id>/formtype/<form_type>", methods=["DELETE"]
+)
+def delete_registry_instrument(exercise_id, form_type):
+
+    if RegistryInstrument().delete_registry_instrument_by_exercise_id_and_formtype(exercise_id, form_type):
```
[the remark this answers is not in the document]
pricem14pc: Done

application/models/models.py
```diff
@@ -189,3 +189,65 @@ def __init__(self, instrument_id=None, file_name=None, length=None, data=None):
...
+    instrument_id = Column(UUID, nullable=False)
+    classifier_type = Column(String, primary_key=True)
+    classifier_value = Column(String, primary_key=True)
+    ci_version = Column(Integer, nullable=False)
+    guid = Column(UUID, nullable=False)
+    published_at = Column(TIMESTAMP, nullable=False)
+
+    def __init__(
+        self,
+        survey_id=None,
```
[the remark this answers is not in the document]
pricem14pc: Good spot ! Hangover from an initial approach before I'd decided these were mandatory.

Done 👍

tests/controllers/test_registry_instrument.py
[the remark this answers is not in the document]
pricem14pc: I mean, I can change the name if you think it doesn't make sense. But the test is literally checking that `get_instruments_by_exercise_id` does indeed `returns_list_of_dicts`

tests/controllers/test_registry_instrument.py
pricem14pc: I looked at that originally but had real trouble with the mock object return value giving me the right thing to assert against.

Or do you mean something like this?

```
expected = {
    "ci_version": ci_version,
    "guid": guid,
    "published_at": published_at,
...
...
blah...
}

actual = {
    "ci_version": mock_registry_instrument_model.return_value.ci_version,
    "guid": mock_registry_instrument_model.return_value.guid,
    "published_at": mock_registry_instrument_model.return_value.published_at,
...
...
blah...
}

self.assertEqual(actual, expected)
```

LJBabbage: Yep, that would work

tests/views/test_registry_instrument_view.py
pricem14pc: Yeah, agree its a bit naff... I'll change it to your suggestion.

tests/views/test_registry_instrument_view.py
pricem14pc: Done

tests/views/test_registry_instrument_view.py
pricem14pc: Can do.. this time I _was_ being lazy !

pricem14pc: Added an assertion of the source object to cover the content of each attribute. 

I admit it now appears overkill, but but going to leave the size assertion in place for visibility of the (very strict) object we're handling. It will also highlight the tests need tweaking if we independently extend the object in the future.

tests/controllers/test_registry_instrument.py
pricem14pc: _think_ I did this one.

application/models/models.py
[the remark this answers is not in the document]
pricem14pc: Actually, this _was_ required based on the way this SQLAlchemy pattern creates the entity (with none values for each field) and then subsequently sets the values. I've reverted.

LJBabbage (approved):
A pragmatic approval. It all works, but it could do with a bit of work in places. The PUT end point is perhaps the area most in need of looking at. It feels like it is doing too much and the name of functions and docs don't align with the journey. ``` _find_or_create``` is actually updating the ci_version, guid and published_at for an existing objects. But the name of function or docs don't say that. It does feel like there should be separation between find or create for future re-use. The PUT process equally has no exception catching even though validate_uuid can raise one and probably creating a new object as well. On validate_uuid(exercise_id) it's done twice in the flow.

Tests are still a tad verbose and cumbersome in places and type hinting would be nice, but all thee above could be picked up by other stories/iterations

matthew-robinson-ons (approved):
Have tested all endpoints and working as expected. Reviewed the comments and happy to approve the current implementation

PR #356, opened 2025-07-31
Interim alembic fix for broken dev envs

pricem14pc:
# What and why?

The existing strategy when creating a new `ras_ci` schema did not consider those objects types NOT defined in SQLAlchemy models. It only considered table objects defined in SQLAlchemy models.

As such, the recent `view` and `index` objects introduced in Alembic migrations for the registry instruments are NOT created when creating a new schema (essentially dev/performance environments). These objects cannot be created by SQLAlchemy definitions in the same way tables are. They can only be created explicitly by code frameworks (i.e. Alembic).

The issue does NOT affect prod or preprod as the schemas are long standing and already exist. As such, the Alembic migrations for those envs are correctly applied. Preprod correctly has the views and index for registry instruments in its schema, along with the correct migration revision `9a724381cde7`.

However, we need to consider the strategies for prod/preprod AND dev environment. These have conflicting requirements.

This PR introduces an inelegant workaround to the problem in dev while I work out a suitable approach. Instead of setting the Alembic revision to the most recent `9a724381cde7` in the assumption that SQLAlchemy has created all the objects, it sets the revision to the previous `b730b3a81f72` and then calls the Alembic script which applies most recent `9a724381cde7` revision containing the registry instrument view (and index).

This is far from ideal, and will need further thinking about when future migrations are added. We might need to update existing migration scripts (which itself not ideal) to check for objects before trying to create them, and then run all the migrations in regardless.

# How to test?

- deploy to a NEW dev env and check the `ras_ci` schema, and all the tables and view and index are created.
- deploy to an EXISTING previous version env, and check the view and index are created.

# Jira

n/a

### DIFF
--- a/migrations/versions/9a724381cde7_add_registry_instrument_count_view.py
+++ b/migrations/versions/9a724381cde7_add_registry_instrument_count_view.py
@@ -20,6 +20,9 @@ def upgrade():
         """CREATE VIEW ras_ci.registry_instrument_count as SELECT exercise_id, count(*) as count
         FROM ras_ci.registry_instrument GROUP BY exercise_id; """
     )
+    op.create_index(
+        "ix_registry_instrument_exercise_id", "registry_instrument", ["exercise_id"], unique=False, schema="ras_ci"
+    )
 
 
 def downgrade():
--- a/run.py
+++ b/run.py
@@ -91,7 +91,11 @@ def create_database(db_connection, db_schema):
             models.Base.metadata.create_all(engine)
             # If the db is created from scratch we don't need to update with alembic,
             # however we do need to record (stamp) the latest version for future migrations
-            command.stamp(alembic_cfg, "head")
+            # THE ABOVE HISTORICAL STATEMENT IS NOT TRUE AND IS NOW CAUSING AN ISSUE. IT DID NOT CONSIDER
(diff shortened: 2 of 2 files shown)


migrations/versions/9a724381cde7_add_registry_instrument_count_view.py
```diff
@@ -20,7 +20,11 @@ def upgrade():
         """CREATE VIEW ras_ci.registry_instrument_count as SELECT exercise_id, count(*) as count
         FROM ras_ci.registry_instrument GROUP BY exercise_id; """
     )
+    op.create_index(
+        "ix_registry_instrument_exercise_id", "registry_instrument", ["exercise_id"], unique=False, schema="ras_ci"
+    )
```
pricem14pc, line 25: We wouldn't normally edit previous migrations, but this is the only way to all us to deal with dev, preprod and prod, all of which are at different revisions. Unlike Java Liquibase, these Alembic migrations do not consider a hash of the migrations files and therefore it's easier to move these statements between migration files. This is a last resort to ensure consistency throughout our envs.

matthew-robinson-ons (approved):
Stepped through the testing to replicate a PROD setup and with a new DEV env, both times all resources were created properly and acceptance tests all passing.

